PR #145, opened 2013-04-08
Fixes #144 treat empty array as constructor args

### LINKED ISSUE #144: Partial mock does not call constructor by default
When building a partial mock via
»Mockery::mock('classname[method]')«
the classname's constructor is not called. 
Even
»Mockery::mock('classname[method]', array())«
does not call the constructor.

But when I set a dummy argument
»Mockery::mock('classname[method]', array('blah'))«
the constructor is called.

Is this intended behaviour?

davedevelopment:


havvg:
This does not work, if the constructor calls methods internally.

```php
// fails as "initialize" is not defined to be mocked.
// (initialize being public method called within constructor)
$request = \Mockery::mock('Symfony\Component\HttpFoundation\Request', []);
$request->shouldDeferMissing();
```

Workaround - at least for the Request class in the example. However, this requires insight of any changes to the class, which is bad.

```php
$request = \Mockery::mock('Symfony\Component\HttpFoundation\Request');
$request->shouldDeferMissing();
$request->initialize();
```

PR #218, opened 2013-11-09
Fixed #215  if we overriding constructor, we override original destructor too

### LINKED ISSUE #215: Overriding original destructor
Hello, 

Say I have a class:

```
class Foo
{
   function __construct() { /* open file */ }
   function __destruct()  { /* close file */ } 
}
```

When i mock this class `$foo = Mockery::mock('Foo');` it override original constructor, but not override destructor. So i had errors in the destructor. 

I think Mockery should override destructor & constructor together. I.e. when i use

`$foo = Mockery::mock('Foo', array());` - it uses original class constructor & destructor
`$foo = Mockery::mock('Foo');` - no original constructor & destructor called

lislon:


neoshadybeat:
:+1:

PR #293, opened 2014-03-03
Mockery::hasSubstrings() matcher

idrarig:
Use as Mockery::hasSubstrings('needed substring', 'another needed substring')

I wrote the NonMatchingCase test in a different way than the other tests, because it seems to me that there's something wrong.  If you change, for instance, the third call to `foo()` in `testHasKeyConstraintNonMatchingCase()`, to `$this->mock->foo(1, array('a'=>1))`, the test still passes; I think it should fail?

### DIFF
--- a/library/Mockery.php
+++ b/library/Mockery.php
@@ -289,6 +289,17 @@ class Mockery
     }
 
     /**
+     * Return instance of HASSUBSTRINGS matcher
+     *
+     * @return
+     */
+    public static function hasSubstrings()
+    {
+        $return = new \Mockery\Matcher\HasSubstrings(func_get_args());
+        return $return;
+    }
+
+    /**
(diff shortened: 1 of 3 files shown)


padraic:
Closing as met by Hamcrest matchers.

@davedevelopment We need a new issue for the test question?

aik099:
Usually GitHub repos don't have ways to ask questions. For that reason it's common for people to post new issues with questions.

Are there any solutions for that except for https://gitter.im/ ?

padraic:
Well, in this case it looks like a question that may technically be a testing bug. Ah, running afoul of the very thing I wrote Humbug for. The irony.

davedevelopment:
In review, I think the test is correct, though the behaviour is vague.

padraic:
It is? Irony avoided!

PR #364, opened 2014-08-13
Not matcher now negate anything

yguedidi:
Now you can use
- ``->with(Mockery::not('/regexp/i))``
- ``->with(Mockery::not('stdClass'))``
- ``->with(Mockery::not(Mockery::type('resource')))``


padraic:
Refer to Hamcrest. If that does not have a suitable matcher, or will not accept a PR to add same, then we can revisit.

PR #378, opened 2014-09-29
Fixed support for creating instances of builtin (eg. SPL) classes in PHP 5.5

asmblah:
Fixes support for mocking eg. Symfony's `UploadedFile` class in PHP 5.5.


padraic:
@davedevelopment Can we close this?

davedevelopment:
Yes, this should be covered in #403

PR #400, opened 2014-12-12
Mockery::equalsStringIgnoringWhitespace() matcher

sk-sparhandy:
EqualsStringIgnoringWhitespace() matches two strings while ignoring whitespace. This is especially useful if you want to match multi line strings which are not indented in the same way.

### DIFF
--- a/library/Mockery.php
+++ b/library/Mockery.php
@@ -382,6 +382,18 @@ class Mockery
     }
 
     /**
+     * Returns an instance of the EQUALSTRINGIGNORINGWHITESPACE matcher.
+     *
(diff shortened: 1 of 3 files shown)


padraic:
Closing - see Hamcrest and perhaps include a PR there if needed. If it does not fulfil the requirements of this matcher we can revisit a solution.

aik099:
We need to place large banner in CONTRIBUTING.md saying: `Issues/PR about new matchers should go directly to Hamcrest` :smile:

padraic:
Await my next RFC :)

PR #415, opened 2015-01-20
NestedSubset matcher

sosso:
Hi,

I needed a more powerful version of the subset matcher for some message queue mocking I was doing, and did not see anything that did it, so I built this!


aik099:
Doesn't Hamcrest have something similar?

padraic:
Closing - see Hamcrest and perhaps include a PR there if needed. If it does not fulfil the requirements of this matcher we can revisit a solution.

PR #417, opened 2015-01-25
Exclude static properties and methods differently when formatting objects.

### LINKED ISSUE #416: Message construction for NoMatchingExpectationException does not handle static properties+methods correctly
`Mockery::extractInstancePublicProperties()` and `Mockery::extractGetters()` use `\ReflectionClass` to retrieve public non-static properties and public non-static methods respectively. Unfortunately the behaviour of filters here isn't as expected, as negation has no effect. https://bugs.php.net/bug.php?id=65044 has more details.

To reproduce:

```php
class Test
{
    public function foo($x) { }
}

class Arg
{
    public static $x;
    public $y;

    public static function getX() { }
    public function getY() { }
}

$m = Mockery::mock('\Test')
    ->shouldReceive('foo')
    ->once()
    ->with(new Arg())
    ->getMock();

$m->foo(new Arg());
```

With the static property included, E_NOTICE and E_STRICT errors are triggered, due to the code retrieving the property in a non-static fashion:

```
PHP Strict Standards: Accessing static property Arg::$x as non static in .../mockery/library/Mockery.php on line 528
PHP Notice: Undefined property: Arg::$x in .../mockery/library/Mockery.php on line 528`
```

The code to retrieve getter methods runs error free. Static properties and getters are erroneously included in the output:

```
No matching handler found for Mockery_0_Test::foo(object(Arg)). Either the method was unexpected or its arguments matched no expected argument list for this method

Objects: ( array (
  'Arg' =>
  array (
    'class' => 'Arg',
    'properties' =>
    array (
      'x' => NULL,
      'y' => NULL,
    ),
    'getters' =>
    array (
      'getX' => NULL,
      'getY' => NULL,
    ),
  ),
))
```

nick-jones:
Assuming the existing *intended* behaviour remains desired, then this resolves #416

### DIFF
--- a/library/Mockery.php
+++ b/library/Mockery.php
@@ -520,12 +520,14 @@ class Mockery
     private static function extractInstancePublicProperties($object, $nesting)
     {
         $reflection = new \ReflectionClass(get_class($object));
-        $properties = $reflection->getProperties(\ReflectionProperty::IS_PUBLIC & ~ \ReflectionProperty::IS_STATIC);
+        $properties = $reflection->getProperties(\ReflectionProperty::IS_PUBLIC);
         $cleanedProperties = array();
 
         foreach ($properties as $publicProperty) {
-            $name = $publicProperty->getName();
-            $cleanedProperties[$name] = self::cleanupNesting($object->$name, $nesting);
+            if (!$publicProperty->isStatic()) {
+                $name = $publicProperty->getName();
+                $cleanedProperties[$name] = self::cleanupNesting($object->$name, $nesting);
+            }
         }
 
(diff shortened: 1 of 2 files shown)


padraic:
@davedevelopment Any comment?

davedevelopment:
I'm a :+1:. I'd really like to replace a lot of this code with [sebastianbergmann/exporter](https://github.com/sebastianbergmann/exporter) et al in the not too distant future.

padraic:
@davedevelopment I have no objections to that.

padraic:
@nick-jones Merged and thanks :P

PR #419, opened 2015-01-30
Support variadic parameters in methods

### LINKED ISSUE #240: PHP 5.6 Variadic functions
Will be a while before we're likely to need it, but the code generation will need to support the [variadics](https://wiki.php.net/rfc/variadics)

duncan3dc:
I'm not that familiar with the internals of Mockey so I may be way off here, but I added a failing test and then attempted an implementation.

Fixes #240 

Let me know if anything needs fixing up

### DIFF
--- a/library/Mockery/Generator/Parameter.php
+++ b/library/Mockery/Generator/Parameter.php
@@ -76,4 +76,16 @@ class Parameter
 
         return $name;
     }
+
+
+    /**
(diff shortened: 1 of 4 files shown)


duncan3dc:
The failing hhvm test appears to be unrelated to my change and currently failing on master

davedevelopment:
This looks like a great start, thank you very much!

PR #421, opened 2015-02-04
TestCase should be abstract

scottrobertson:
Otherwise, sometime phpunit will fail due to no tests being found in MockeryTestCase

### DIFF
--- a/library/Mockery/Adapter/Phpunit/MockeryTestCase.php
+++ b/library/Mockery/Adapter/Phpunit/MockeryTestCase.php
@@ -4,7 +4,7 @@ namespace Mockery\Adapter\Phpunit;
 
 use Mockery;
 
-class MockeryTestCase extends \PHPUnit_Framework_TestCase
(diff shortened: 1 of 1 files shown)


aik099:
I guess `sometimes` refers to the case, when test file suffix in `phpunit.xml` is set to `TestCase.php` vs default `Test.php`.

scottrobertson:
Hmm, we get this currently, and our suffix is Test.php

aik099:
How it's even possible, that PHPUnit is able to look into `/vendor` folder while it looks for possible projects tests.

Can you please post your `phpunit.xml`?

scottrobertson:
It's not, we are using this in the following way:

```
abstract class AbstractTestCase extends MockeryTestCase
```

It's very odd, it only seems to do it if we run phpunit on a whole directory, and not individual files, or just from the root.

aik099:
You're absolutely right in making the class abstract, because empty (without tests) test case is not of much use anyway.

davedevelopment:
Thanks!

PR #420, opened 2015-02-02
Add 'addRiskyTest' to maintain abstract class relationship in TestListener

dustindoiron:
Fixes PHPUnit 4.4 incompatibility.

Tested on PHPUnit 4.4.5 and PHPUnit 3.7.28.

PR #422, opened 2015-02-05
Update TestListener.php

Varell09:
addRiskyTest war declared too times - exception throwed because of redeclaration.


aik099:
That is really interesting because #420 PR author, that was merged yesterday saw that there were no such method.

davedevelopment:
Doh! Thanks.

WouterJ:
@aik099 in #420 , there already was such method (click on the fold icon to see the lines above the added line, including one with this method). The author was probably using an older version of Mockery

PR #425, opened 2015-02-10
Format null argument correctly

robertbasic:
When an expectation is set for a string, but a null is passed instead,
make sure the argument is correctly formatted in the exception message
so we don't get an empty string printed where null should be.

Ended up chasing my tail a couple of times because of messages like `No matching handler found for Mockery_154_Foo::foo("")` :)

### DIFF
--- a/library/Mockery.php
+++ b/library/Mockery.php
@@ -447,6 +447,10 @@ class Mockery
             return 'resource(...)';
         }
 
+        if (is_null($argument)) {
+            return 'NULL';
+        }
+
         $argument = (string) $argument;
 
         return $depth === 0 ? '"' . $argument . '"' : $argument;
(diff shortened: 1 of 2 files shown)


aik099:
:+1:

aik099:
I used to place breakpoint in expectation matching method in Mockery to determine what actually wasn't matched too.

robertbasic:
Now all we need is array diffs like phpunit has.

/me runs

davedevelopment:
Thanks @robertbasic 

The diffs shouldn't be too difficult, between sebastianbergmann/exporter, sebastianbergmann/comparator and sebastianbergmann/diff. @karptonite and I were going to see about trying to do it in a way that didn't change too much, but got rid of most of the mockery code that does the same thing, but I don't think it's worth it, we should just wholesale replace it.

davedevelopment:
*change too much* in the way it looks

Issue #424, opened 2015-02-10
Mockery does not provide integration with PHPUnit


Bilge:
PHPUnit has a test mode designed to "[report useless tests](https://phpunit.de/manual/current/en/strict-mode.html)" which marks tests that do not perform any assertions or expectations. However, PHPUnit is ignorant to any expectations performed on Mockery objects since Mockery has no integration with PHPUnit.

davedevelopment:
We have the [MockeryTestCase](https://github.com/padraic/mockery/blob/master/library/Mockery/Adapter/Phpunit/MockeryTestCase.php) which I believe helps with this

aik099:
@davedevelopment is correct. Due restrictions of PHPUnit test listeners (which default Mockery integration mode with PHPUnit) it's unable to properly report such tests.

Bilge:
Wouldn't it be better implemented as a trait?

aik099:
There is a trait already if I'm not mistaken, but projects which use PHP 5.3 are unable to use it and that's why there is also abstract test case class.

davedevelopment:
There's no trait at this time, we're still supporting 5.3 and I'm not overly keen on the idea of using traits for this kind of extension.

Bilge:
Introducing an optional trait does not relinquish support for 5.3.

davedevelopment:
@Bilge, no but we'd need to copy-paste any changes to both the trait and the abstract class in the mean time. I'm not against a trait, but I personally wouldn't use it

Bilge:
That doesn't surprise me. From the way you're acting you've never harnessed horizontal reuse in your life.

davedevelopment:
@Bilge if you would like a trait, please submit a PR and I'll happily merge it, but please don't insult me, thanks.

Bilge:
I just wanted you to be aware of your xenophobic attitude.

davedevelopment:
@Bilge okay, thanks.

Bilge:
It was never my intention to insult you :+1:

I'm considering submitting a PR soon, but even if it were accepted, this issue would not be resolved. My impression was that there was no attempt made to solve this issue, and yet there is, which is an outstanding documentation issue unless you can show me the part of the documentation that describes this class that I missed.

davedevelopment:
No, there are no docs yet, the only acknowledgement of it's existence is in the changelog. I'm well behind on the docs. Labelled this one as a documentation issue and look forward to your PR.

[...]

PR #428, opened 2015-02-11
Added MockeryPHPUnitIntegration.

Bilge:
Re: #424.

### DIFF
--- a/library/Mockery/Adapter/Phpunit/MockeryPHPUnitIntegration.php
+++ b/library/Mockery/Adapter/Phpunit/MockeryPHPUnitIntegration.php
@@ -1,0 +1,27 @@
+<?php
+namespace Mockery\Adapter\Phpunit;
+
+/**
+ * Integrates Mockery into PHPUnit. Ensures Mockery expectations are verified
+ * for each test and are included by the assertion counter.
+ */
+trait MockeryPHPUnitIntegration
+{
+    /**
+     * Performs assertions shared by all tests of a test case. This method is
+     * called before execution of a test ends and before the tearDown method.
+     */
+    protected function assertPostConditions()
+    {
+        parent::assertPostConditions();
+
+        // Add Mockery expectations to assertion count.
+        if (($container = \Mockery::getContainer()) !== null) {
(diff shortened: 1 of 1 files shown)


padraic:
@Bilge Could we impose on you to include something about this in the docs also? :)

Bilge:
I don't think so, bro. I'd love to but I don't have that kind of time.

padraic:
@Bilge No problem.

@davedevelopment I'll let you merge and then you can ping me to write some docs.

davedevelopment:
Thanks!

fhinkel:
This breaks the 5.3 build (for obvious reasons).

aik099:
From build log I don't get which test calls the trait. I guess it's PHPUnit itself scans all files and hits the error.

aik099:
It seems, that PHUnit tried to build coverage report and since file with a trait is within whitelisted folder it includes it and gets an error.

Maybe we should update `phpunit.xml` to exclude file with trait from ones, requiring coverage checks.

padraic:
There's an easy fix for that ;).

I tried a blacklist, but no joy. Would need to remove the uncoveredfiles attribute on the whitelist I think.

padraic:
@fhinkel @aik099 Fixed in current master. Just excluded it from whitelist for now.

aik099:
Thanks.

Travis badge still red, but this is due HHVM eval problem, which suggests that we either need to fix this or mark HHVM tests are allow-failure.

davedevelopment:
Sorry, didn't even think about the coverage reports etc

padraic:
@aik099 I'll look at the HHVM reports later. I also did some checking and it looks like HHVM pins to specific Mockery version for its framework testing. I might be reading it wrong but they are pinned to 0.9.2. Anyways, if we can fix them, I will though at some point HHVM can get its own stuff in order if a given fix is looking a bit nuts.

fhinkel:
@padraic thank you!

padraic:
Note: Master will now undo the trait include for code coverage for PHP 5.6 builds only using a small sed delete of that line.

Issue #391, opened 2014-11-09
Add 'shouldNotReceive' expectation


mlessnau:
I think it would be consistent to have a 'shouldNotReceive' expectation.

Currently, the way to go is writing ```$m->shouldReceive('foo')->never()```. This could then be written as ```$m->shouldNotReceive('foo')```.

padraic:
I have no real objections. @davedevelopment ?

robertbasic:
:+1:

aik099:
:+1:

davedevelopment:
Fixed in #391

PR #430, opened 2015-02-13
Feature/issue 391 add should not receive expectation

robertbasic:
Took a stab at #391

### DIFF
--- a/docs/reference/expectations.rst
+++ b/docs/reference/expectations.rst
@@ -45,7 +45,8 @@ usage (e.g. during refactoring). See examples in a later section.
 
     shouldNotReceive(method_name)
 
-Declares that the mock should not expect a call to the given method name.
+Declares that the mock should not expect a call to the given method name. This
+method is a convenience method for calling ``shouldReceive()->never()``.
 
 .. code-block:: php
 
--- a/docs/reference/expectations.rst
+++ b/docs/reference/expectations.rst
@@ -43,6 +43,12 @@ usage (e.g. during refactoring). See examples in a later section.
 
 .. code-block:: php
 
+    shouldNotReceive(method_name)
+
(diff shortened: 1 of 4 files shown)


library/Mockery/Mock.php
```diff
@@ -203,6 +203,19 @@ public function shouldReceive()
     }
 
     /**
+     * Shortcut method for setting an expectation that a method should not be called.
+     *
+     * @param mixed
+     * @return \Mockery\Expectation
+     */
+    public function shouldNotReceive()
```
aik099: I guess this needs to be added to `MockInterface` as well.

docs/reference/expectations.rst
```diff
@@ -43,6 +43,12 @@ usage (e.g. during refactoring). See examples in a later section.
 
 .. code-block:: php
 
+    shouldNotReceive(method_name)
+
+Declares that the mock should not expect a call to the given method name.
```
davedevelopment: I think a little comment here to say that it is effectively shorthand for `shouldReceive()->never()` would be good.

padraic: :+1:

robertbasic:
@aik099 @padraic @davedevelopment PR updated!

tests/Mockery/ExpectationTest.php
```diff
@@ -402,6 +402,31 @@ public function testNeverCalled()
         $this->mock->shouldReceive('foo')->never();
     }
 
+    public function testShouldNotReceive()
+    {
+        $this->mock->shouldNotReceive('foo');
```
aik099: Should you be calling `$this->container->mockery_verify();` here as well and verify, that if more than 0 calls are made we're getting an exception?

davedevelopment:
Good work @robertbasic

Issue #177, opened 2013-06-17
shouldIgnoreMissing() not working on instance mocks


david0:
`shouldIgnoreMissing()` does not work on mock objects created using instance mocks.

```php
$m = m::mock('overload:A');
$m->shouldIgnoreMissing();

$a = new A();
$a->doSomething();
```
leads to 
`BadMethodCallException: Method ::doSomething() does not exist on this mock object`

I think it this function would be pretty useful, especially for instance mock objects. Also it looks a little inconsistent, since one can set expectations using `shouldReceive` on all instances, but `shouldIgnoreMissing` only affects the current mock object.

robertbasic:
`shouldIgnoreMissing` and `shouldDeferMissing` (and other such flags present and future) do work, but not on the mock, but on the instance mock.

Sounds confusing, but it's because of internals.

When you do a `m::mock('overload:Foo');` mockery goes and creates a new class called `Foo` based on the `Mock.php` "template". And that template has the `_mockery_ignoreMissing` flag set to `false` by default.

Once the `mock` method gets to the end, it creates an overloaded `Foo` object and returns it to us. The overloaded `Foo` class is also available to instantiate new `Foo` objects.

Calling `$m->shouldIgnoreMissing()` sets the ignore missing flag on the object that was returned by the `mock` method.

Calling `new Foo()` creates a new instance of the `Foo` object that has the `Mock.php` templates default of `false` for `_mockery_ignoreMissing`.

If you want the flag set to `true` on the new instance, you need to call `shouldIgnoreMissing()` on that instance, and not on `$m`.

Obviously this is far from perfect.

One idea I have is to have a global configuration for this, I have a proof of concept in this commit https://github.com/robertbasic/mockery/commit/32d70b7463a83bcac7338a1bc6a0d5a6da1cb469

Thoughts?

/ping @padraic

padraic:
@davedevelopment Thoughts?

See how well I can delegate? :laughing:

davedevelopment:
When you do a `new ClassThatHasBeenOverloaded`, we have a [miracle constructor](https://github.com/padraic/mockery/blob/master/library/Mockery/Generator/StringManipulation/Pass/InstanceMockPass.php#L30) that copies the expectations over from the `master` object, we should probably copy over the relevant flags as well?

robertbasic:
Hm I was looking in there, but didn't see anything useful. Will try again that then.

davedevelopment:
[...]

PR #436, opened 2015-02-13
Feature/issue 177 should ignore missing on instance mocks

robertbasic:
For #177

@davedevelopment's idea works/looks way better than mine!


davedevelopment:
Let's see how it goes :)

Issue #396, opened 2014-11-14
namedMock does not use global namespace in function parameter type hints


nigelchanyk:
``` php
<?php

namespace Foo {

  class Bar {
    public function value() {
      return 0;
    }
  }

}

namespace {

  class Meh {
    public function doSomething(\Foo\Bar $bar) {
      return $bar->value();
    }
  }

  $meh = \Mockery::namedMock("ILoveRandomNamespace\\Meh", "Meh");
  $bar = new \Foo\Bar();
  $meh->shouldReceive('doSomething')->once()->andReturn(1337);
  var_dump($meh->doSomething($bar));
}
```

Error:
Catchable fatal error: Argument 1 passed to ILoveRandomNamespace\Meh::doSomething() must be an instance of ILoveRandomNamespace\Foo\Bar, instance of Foo\Bar given, called in /home/nchan/meh.php on line 24 and defined in /data/users/fbcode/hphp/test/frameworks/framework_downloads/mockery/library/Mockery/Loader/EvalLoader.php(16) : eval()'d code on line 704

Cause:
The eval()'d code has namespace ILoveRandomNamespace. The generated doSomething() function in the eval()'d code accepts parameter Foo\Bar. Since we're still in ILoveRandomNamespace, ILoveRandomNamespace\Foo\Bar is inferred.

robertbasic:
Assigning to myself as I tried a PR for this a long time ago...

robertbasic:
Same as #525, this issue was fixed with PR #558 

Closing as fixed.

PR #437, opened 2015-02-14
Prepend namespace separator to type hinted class

robertbasic:
ReflectionParameter gets typehinted classes with
fully resolved names, but does not prepend the
namespace separator to it.

See issue #396

### DIFF
--- a/library/Mockery/Generator/Parameter.php
+++ b/library/Mockery/Generator/Parameter.php
@@ -48,7 +48,13 @@ class Parameter
         if ((version_compare(PHP_VERSION, '5.4.1') >= 0)) {
             try {
                 if ($this->rfp->getClass()) {
-                    return $this->rfp->getClass()->getName();
+                    $typehintedClass = $this->rfp->getClass()->getName();
+
+                    if ($this->isNamespacedClass($typehintedClass)) {
+                        return '\\' . $typehintedClass;
+                    }
+
+                    return $typehintedClass;
                 }
             } catch (\ReflectionException $re) {
                 // noop
@@ -56,7 +62,13 @@ class Parameter
         }
 
         if (preg_match('/^Parameter #[0-9]+ \[ \<(required|optional)\> (?<typehint>\S+ )?.*\$' . $this->rfp->getName() . ' .*\]$/', $this->rfp->__toString(), $typehintMatch)) {
+
             if (!empty($typehintMatch['typehint'])) {
+
+                if ($this->isNamespacedClass($typehintMatch['typehint'])) {
+                    return '\\' . $typehintMatch['typehint'];
+                }
+
                 return $typehintMatch['typehint'];
             }
         }
@@ -87,5 +99,13 @@ class Parameter
             return false;
         }
         return $this->rfp->isVariadic();
+    }
+
+    private function isNamespacedClass($class)
+    {
+        if (strpos($class, '\\') > 0) {
+            return true;
+        }
+        return false;
     }
 }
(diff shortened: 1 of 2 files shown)


robertbasic:
Sorry about the ugly tests, any ideas how to write that more nicely?

Also, I think it should be safe/better to prepend the `\\` to all typehinted classes?

library/Mockery/Generator/Parameter.php
```diff
@@ -48,7 +48,13 @@ public function getTypeHintAsString()
         if ((version_compare(PHP_VERSION, '5.4.1') >= 0)) {
             try {
                 if ($this->rfp->getClass()) {
-                    return $this->rfp->getClass()->getName();
+                    $typehintedClass = $this->rfp->getClass()->getName();
+
+                    if ($this->isNamespacedClass($typehintedClass)) {
```
aik099: The IF isn't necessary, because prepending `\` before class name won't do any harm for classes in global namespace.

library/Mockery/Generator/Parameter.php
```diff
@@ -56,7 +62,13 @@ public function getTypeHintAsString()
         }
 
         if (preg_match('/^Parameter #[0-9]+ \[ \<(required|optional)\> (?<typehint>\S+ )?.*\$' . $this->rfp->getName() . ' .*\]$/', $this->rfp->__toString(), $typehintMatch)) {
+
             if (!empty($typehintMatch['typehint'])) {
+
+                if ($this->isNamespacedClass($typehintMatch['typehint'])) {
```
aik099: Same (if removal) here.

library/Mockery/Generator/Parameter.php
robertbasic: I was thinking the same. Reflection always gets the FQN of the typehinted class.

aik099: Reflections always give you back FQCN class name without `\`. As you've noted with used in `eval` this ends up pretty badly by inheriting namespace of current class, where eval is made.

library/Mockery/Generator/Parameter.php
robertbasic: I think some kind of a check needs to be in place here, for the case when `typehint` is not a class name, no? For example `callable`.

aik099: Currently the only ofther thing could be interface name or `array`/`callable` word. If the scalar type typehints RFC will be accepted, then there could be more.

I don't know any method that can tell that given string is one of build-in types.

Maybe also typehints can be `self` or `static`, but I'm not sure at 100%.

robertbasic: We can call `get_declared_classes` and check the preg_matched typehint against those.

aik099: Ha, look few lines above in this class. There is already such code, that I guess needs to be extracted in separate method:
typehinttextmethod

robertbasic: I think that's a HHVM method, it's not available for "normal" PHP

GuillaumeDievart:
Then ?

library/Mockery/Generator/Parameter.php
aik099: @robertbasic , please proceed with creating/using method that determines if it's an internal type (e.g. array/callable) to avoid adding `\\` before them. Also add tests for such case.

davedevelopment:
I don't think this is still applicable...

PR #438, opened 2015-02-15
Cleanup Travis builds and implement PHP7 target

padraic:
This does a few things:

* Extracts all the scripts out to separate bash scripts for easier editing.
* It uses the 5.3.3 build target to compile and use php-src master (i.e. PHP 7)
* Fixes the mongo and redis extension loading
* Adds a composer self-update where needed
* For the PHP 7 (aka 5.3.3) build, installs composer manually
* Code coverage and scrutinizer will now target PHP 5.6 (was 5.5)
* Move satooshi/php-coveralls from composer.json to travis composer require command

Note: A previous commit moved hhvm to allowed_failures so PRs can pass checks. The 5.3.3 (actually PHP 7) is currently not allowed to fail.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ php:
   - 5.4
   - 5.5
   - 5.6
+  - 5.3.3 #This will actually use PHP7
   - hhvm
   - hhvm-nightly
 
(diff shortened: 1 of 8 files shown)


padraic:
Code coverage seems to fall under 5.6 slightly but I think it's likely just version specific code being eliminated since no longer running on 5.5.

padraic:
Code coverage reduction appears to arise from using PHP 5.6 over 5.5 for generating coverage. The changes mean we're not generating for every single PHP version.

davedevelopment:
Awesome!

PR #439, opened 2015-02-15
CS fixes to PSR-2 level

padraic:


aik099:
I hope you haven't done all these fixes by hand. There is PHP_CodeSniffer 2.x versions, that not only reports issues, but also can fix them. And there is PHP-CS-Fixer.

GrahamCampbell:
And https://styleci.io/ launches soon. :)

padraic:
@aik099 @GrahamCampbell php-cs-fixer :laughing: 

@GrahamCampbell Very nice! I'll be sure to check it out when it launches.

GrahamCampbell:
We're using php-cs-fixer under the hood, so we'll respect any .php_cs file with your config in.

GrahamCampbell:
@padraic Just to let you know that I've launched https://styleci.io/ now. :)

aik099:
That's great news. Some more docs wouldn't hurt in there. I've tried to use it and the only thing I was able to do is to get list of my repos and enable StyleCI on one of them. I have no idea what will happen and where do I specify standard to be checked. Actually I have no idea what that website does in general.

PR #444, opened 2015-02-26
unused namespace removed

TomasVotruba:


padraic:
Thanks @TomasVotruba !

suwardany:
This is breaking the `EvalLoader`

```
Fatal error: Interface 'MockInterface' not found in ***/vendor/mockery/mockery/library/Mockery/Loader/EvalLoader.php(16) : eval()'d code on line 24
```

GrahamCampbell:
#449

TomasVotruba:
Would not be better to fix EvalLoader, so it doesn't require this custom behavior? It hidden feature.

PR #446, opened 2015-02-26
Add options to make all methods as stubs, returning null by default

yeka:
I'm referring to this documentation of PHPUnit for creating mocked object that all methods are stubs, which return null by default: [Unit Testing Tutorial](https://jtreminio.com/2013/03/unit-testing-tutorial-part-5-mock-methods-and-overriding-constructors/) (Do not call setMethods() & Passing an empty array)

This kind of feature doesn't seems currently available on Mockery.

One option that close to that is `shouldIgnoreMissing()`, but this option will always return null even when the method being called is not exists. This is not good when there's a typo in either unit test's class or actual class.

The additional option of `makeStubs()` works alike, but will still throw error when a nonexistent method is called.

### DIFF
--- a/library/Mockery/MockInterface.php
+++ b/library/Mockery/MockInterface.php
@@ -81,6 +81,13 @@ interface MockInterface
     public function makePartial();
 
     /**
+     * Set mock to return null by default for all existing methods
+     *
+     * @return Mock
+     */
+    public function makeStubs();
+
+    /**
      * @param $method
      * @param null $args
      * @return \Mockery\Expectation
--- a/library/Mockery/MockInterface.php
+++ b/library/Mockery/MockInterface.php
@@ -88,6 +88,14 @@ interface MockInterface
     public function disallowMockingNonExistentMethods();
 
     /**
+     * Return flag indicating whether mocking non-existent methods allowed
+     * Fallback to global configuration
+     *
+     * @return bool
+     */
+    public function mockingNonExistentMethodsAllowed();
+
+    /**
      * @param $method
      * @param null $args
      * @return \Mockery\Expectation
--- a/library/Mockery/MockInterface.php
+++ b/library/Mockery/MockInterface.php
@@ -81,11 +81,11 @@ interface MockInterface
     public function makePartial();
 
     /**
-     * Set mock to return null by default for all existing methods
+     * Disable mocking non-existent methods
      *
      * @return Mock
      */
-    public function makeStubs();
+    public function disallowMockingNonExistentMethods();
 
     /**
      * @param $method
--- a/library/Mockery.php
+++ b/library/Mockery.php
@@ -651,7 +651,7 @@ class Mockery
         $methodNames = explode('->', $arg);
         reset($methodNames);
 
-        if (!\Mockery::getConfiguration()->mockingNonExistentMethodsAllowed()
+        if (!$mock->mockingNonExistentMethodsAllowed()
             && !$mock->mockery_isAnonymous()
             && !in_array(current($methodNames), $mock->mockery_getMockableMethods())
         ) {
--- a/library/Mockery.php
+++ b/library/Mockery.php
@@ -86,6 +86,17 @@ class Mockery
     /**
      * @return \Mockery\MockInterface
      */
+    public static function stub()
+    {
+        $args = func_get_args();
+        return call_user_func_array(array(self::getContainer(), 'mock'), $args)->shouldIgnoreMissing()
+            ->disallowMockingNonExistentMethods()
+            ;
+    }
+
+    /**
+     * @return \Mockery\MockInterface
+     */
     public static function instanceMock()
     {
         $args = func_get_args();
--- a/library/Mockery/Mock.php
+++ b/library/Mockery/Mock.php
@@ -163,7 +163,7 @@ class Mock implements MockInterface
             $this->_mockery_partial = $partialObject;
         }
 
-        if (!\Mockery::getConfiguration()->mockingNonExistentMethodsAllowed()) {
+        if (!$this->mockingNonExistentMethodsAllowed()) {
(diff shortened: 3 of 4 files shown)


davedevelopment:
So we have the second part you desire as a global flag `Mockery::getConfiguration()->allowMockingNonExistentMethods()`, if you really only want it on a per mock basis, perhaps we should make that as a separate method available on the individual mocks.

``` php
$foo = m::mock("Foo")
    ->shouldIgnoreMissing()
    ->disallowMockingNonExistentMethods();
```

ykguru:
Neither work as I expected (couldn't find `disallowMockingNonExistentMethods()` btw). 
Lets say here's my Foo class:
```php
class Foo {
    public function bar() {
        return 'bar';
    }
}
```

Using `Mockery::getConfiguration()->allowMockingNonExistentMethods(false)`, when calling `$foo->bar()` it returns `null` (as expected). But the problem is, I can call `$foo->ghost()` and it also returns `null`. I was expecting an error because `Foo` doesn't have `ghost()` method.
Or is this a bug?

ykguru:
Sorry about the `disallowMockingNonExistentMethods()`, I failed to see that it was a suggestion :p

davedevelopment:
Ah, it seems the `shouldIgnoreMissing` doesn't respect the `allowMockingNonExistentMethods`. I'd say it was a bug, but I guess you could argue either way.

ykguru:
On my latest commit, I got that part covered.
I've also add some unit tests for that.

GrahamCampbell:
Needs rebasing so we're running with the new test setup.

yeka:
Rebased

GrahamCampbell:
Please fix the cs issues as shown by StyleCI.

GrahamCampbell:
You can download the diff by clicking the details link, and apply with using git if you want, or you can just manually apply the fixes.

library/Mockery/MockInterface.php
```diff
@@ -81,6 +81,21 @@ public function shouldDeferMissing();
     public function makePartial();
 
     /**
+     * Disable mocking non-existent methods
+     *
+     * @return Mock
+     */
+    public function disallowMockingNonExistentMethods();
```
GrahamCampbell: Adding methods to an interface is a breaking change btw. We'd need a 0.10 release for this.

davedevelopment: We don't *need* to, there's no BC guarantee until 1.0 IMO

yeka: @GrahamCampbell I got your point.
The `disallowMockingNonExistentMethods()` can be removed from the interface and become a unique method for `Mockery\Mock` class.

But for `Mock` class to be able to locally (per mock basis, without setting it from global configuration) disallow mocking non existent methods, the `Mockery` class needs to have a way to read the local flag. Thus, adding `mockingNonExistentMethodsAllowed()` to the `MockInterface` is inevitable.

I'll try to look for a better way implementing this. Hopefully I can find one that doesn't need to change the interface :D

yeka:
I think the `disallowMockingNonExistentMethods()` for individual mock can wait.
I create another pull request to fix the issue related with global configuration.
`shouldIgnoreMissing()` and `Mockery::getConfiguration()->allowMockingNonExistentMethods()` should now work as a team

padraic:
When I see extra method calls, I start thinking we should simplify. Mockery::stub()? Should we be making this a first class citizen?

yeka:
I'm not really sure how you're gonna make Mockery::stub() works, is it the same like Mockery::mock()?
On my works, the need for stubs is more common than mocking a non existent methods.
If you could elaborate your thoughts on implementing this Mockery::stub(), I would be more than happy to help write the code :D

padraic:
@yeka Merely a thought. I'm not sure how possible or not it is without checking.

ykguru:
Just a thought, how about making mock object as stubs by default?
So when we mock an object, rather than not be able to call its method because we haven't define its behavior, instead it will return `null` by default (or any defined default return value).
Let me know what you think and concern (if any) regarding this.

davedevelopment:
Do you mean like `Mockery::spy()`?

yeka:
No, `Mockery::spy()` is a shortcut for `shouldIgnoreMissing()`.
My thought was changing the default behavior of `Mockery::mock()` so with that line alone, we would be able to call existing function and return `null` by default.

Given that my `Foo` class has a method `bar()`

Current behavior:
```php
$foo = Mockery::mock('Foo');
$foo->bar();
```
This will result in `BadMethodCallException`

Suggested default behavior:
`$foo->bar()` will return `null` instead of exception.

I don't know if many people make use of current behavior. But I'm just telling the idea anyway.

davedevelopment:
IMO, that wouldn't be a mock any more it would be a spy, that's why we have `Mockery::spy()`. It looks to me like your desired behaviour would be solved by using  `Mockery::getConfiguration()->allowMockingNonExistentMethods(false)` in your bootstrap and `Mockery::spy()`. 

I'm not keen on only being able to stub/mock existing methods, as it makes interface discovery more difficult for me, instead I prefer to use `Mockery::getConfiguration()->allowMockingNonExistentMethods(false)` on a CI run, when I've finished with the design phase of TDD.

yeka:
Understood.
Yeah, my desired behavior should be solved by using `Mockery::getConfiguration()->allowMockingNonExistentMethods(false)` and `Mockery::spy()`, except that it doesn't work as expected.

Btw, could you give some thought to my another [pull request](https://github.com/padraic/mockery/pull/450)?

davedevelopment:
Sorry, I should have said "once your PR gets merged", just :+1:'d that PR, will let @padraic have his say but I'm ready to merge.

yeka:
Thanks, I'm a bit relaxed now :p

PR #447, opened 2015-02-26
CS Fixes

GrahamCampbell:
We've shipped https://blog.styleci.io/redefining-configuration/.

### DIFF
--- a/.styleci.yml
+++ b/.styleci.yml
@@ -1,6 +1,10 @@
 preset: recommended
 
+enabled:
+  - concat_with_spaces
+
 disabled:
   - short_array_syntax
+  - phpdoc_no_package
   - new_with_braces
-  
+  - concat_without_spaces
--- a/.styleci.yml
+++ b/.styleci.yml
@@ -1,0 +1,6 @@
+preset: recommended
+
+disabled:
+  - short_array_syntax
+  - new_with_braces
+  
--- a/examples/starship/Starship.php
+++ b/examples/starship/Starship.php
@@ -2,7 +2,6 @@
 
 class Starship
 {
-
     protected $_engineering = null;
 
     public function __construct($engineering)
@@ -21,5 +20,4 @@ class Starship
         // We can add more runDiagnosticLevel() calls without failing the test
         // anywhere above since they are unordered.
     }
-
 }
--- a/examples/starship/StarshipTest.php
+++ b/examples/starship/StarshipTest.php
@@ -18,5 +18,4 @@ class StarshipTest extends PHPUnit_Framework_TestCase
         $starship = new Starship($mock);
         $starship->enterOrbit();
     }
-
 }
--- a/library/Mockery.php
+++ b/library/Mockery.php
@@ -1,6 +1,6 @@
 <?php
 /**
- * Mockery
+ * Mockery.
  *
  * LICENSE
  *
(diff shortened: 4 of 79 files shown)


GrahamCampbell:
What's up with scrutinizer?

padraic:
Level should be PSR2 (if different from recommended). It's not actually established policy per se so... @davedevelopment Thoughts?

As a general comment, I'd also prefer to see recommended called something more neutral so people can put their flak cannons away. PSR-2 is also recommended - apparently by me since I'm a PHP-FIG voting rep ;). I'm all for evolving PSR-2, but the appetite for that isn't quite there at the moment.

davedevelopment:
In all honesty, I prefer to wing it, but cave to the standards for standards sake. I try not to pay attention because things like this just make me wonder about wasted energy :)

``` php
-interface TestTraversableInterface extends \Traversable {}
-interface TestTraversableInterface2 extends \Traversable, \Iterator {}
-interface TestTraversableInterface3 extends \Traversable, \IteratorAggregate {}
+interface TestTraversableInterface extends \Traversable
+{
+}
+interface TestTraversableInterface2 extends \Traversable, \Iterator
+{
+}
+interface TestTraversableInterface3 extends \Traversable, \IteratorAggregate
+{
+}
```

davedevelopment:
So PSR-X is fine by me

GrahamCampbell:
@padraic StyleCI's recommended preset basically just extends PSR-2. It includes PSR-2's rules, and adds some extras from symfony.

GrahamCampbell:
@padraic The "recommended" preset is also aliased to "styleci", so if you don't like "recommended", you can use that instead. Also, I wouldn't recommend psr2 on it's own, I'd recommend our recommended preset because it brings you a load of rules that were forgotten about in psr2, or where intentionally avoided when the spec was designed.

padraic:
@GrahamCampbell While I appreciate PSR-2 is imperfect, it's nevertheless the formal recommendation I adhere to for consistency across all projects where possible. Ye olde brain can only handle one standard at a time. Another recommended standard would need a clear specification somewhere to refer to, and a name to go with it, and some community backing as a standard, before I could consider adopting it and asking folk submitting PRs to adhere to it.

That's my primary concern - it's fine to quietly use some extra well understood rules in private if they pass PSR-2 checks in public (or rather the subset covered by php-cs-fixer), but something different when contributors are left confused about enforced extras on their PRs, and what exactly specifies them, and what they're even collectively called in the wild, or even who is recommending them. I can't put that on contributors as an additional burden when they could easily pass PSR-2 or perhaps even symfony, but still get CI failures out of step with other projects.

So, while having no negative opinion at all on what may well be an improved recommendation, Mockery will stick to the path of least resistance on this particular issue. If it does gain traction, it's backup up by documentation, and my concerns evaporate, then we can revisit my and Dave's choice knowing that there is a really nice CI platform (and I really do think it's excellent!) out there to support us.

I didn't really want to turn this into a StyleCI commentary ;), I can do that separately once I've been using it for a while and post some initial thoughts elsewhere closer to your domain.

padraic:
@Scrutinizer You're bugging me! I'm beginning to think the round of 5.3.3 is somehow leaking into how Scrutinizer times out on waiting for code coverage. The PR to use the new 7.0 nightly builds should resolve that so we'll see.

GrahamCampbell:
@padraic I can resend this just with the psr2 preset enabled if you'd really like?

padraic:
@GrahamCampbell That would be my and Dave's consensus for now. I'm with Dave in not being too picky about adhering to a standard, but the consistency is helpful in the long term. I've seen code here that I've written which I could strangle myself for ;).

GrahamCampbell:
Alright then, here's another pull then: https://github.com/padraic/mockery/pull/448.

PR #449, opened 2015-02-27
Revert "unused namespace removed"

GrahamCampbell:
This isn't actually unused.

---
Reverts padraic/mockery#444


padraic:
Noted. Totally my bad.

PR #454, opened 2015-03-10
by ref & should be before variadic ...

williambailey:
As per https://wiki.php.net/rfc/variadics#by-reference_capture by-reference variadic should be in the form of `&...$params`


aik099:
I guess without this you're getting fatal error.

williambailey:
Without this I get `PHP Parse error:  syntax error, unexpected '&', expecting variable (T_VARIABLE) in /vendor/mockery/mockery/library/Mockery/Loader/EvalLoader.php(16) : eval()'d code on line` followed by `PHP Fatal error:  Class 'Mockery_abcd' not found in /vendor/mockery/mockery/library/Mockery/Container.php(470) : eval()'d code on line 1`.
Running PHP 5.6.6

padraic:
Thanks for the PR!

PR #462, opened 2015-04-01
Prep a 0.9.4 release

davedevelopment:
Someone asked for a tag in the irc channel, suggestions and improvements for the changelog welcome


padraic:
We should probably note that this is the final version to support PHP 5.3. Once it's tagged, PHP 5.4 becomes the minimum version supported (and will be reflected in composer.json and the travis conf).

davedevelopment:
@padraic done :)

PR #470, opened 2015-05-11
Add varargs to Mockinterface::shouldReceive and MockInterface::shouldNotReceive

fgrosse:
I stumbled upon this after upgrading PhpStorm to version 9 (early access program).
It does currently mark all calls to `shouldReceive` and `shouldNotReceive`  with the following warning:

> Method call uses 1 parameters, but method signature uses 0 parameters

Looking at the `MockInterface` is also somewhat confusing since it does not define any parameters whatsoever. I am not sure why this is the case but I believe making them explicit at the interface is the least confusing solution.

### DIFF
--- a/library/Mockery/Mock.php
+++ b/library/Mockery/Mock.php
@@ -168,10 +168,10 @@ class Mock implements MockInterface
     /**
      * Set expected method calls
      *
-     * @param mixed
+     * @param mixed ...
      * @return \Mockery\Expectation
      */
-    public function shouldReceive()
+    public function shouldReceive($args)
     {
         /** @var array $nonPublicMethods */
         $nonPublicMethods = $this->getNonPublicMethods();
@@ -207,10 +207,10 @@ class Mock implements MockInterface
     /**
      * Shortcut method for setting an expectation that a method should not be called.
      *
-     * @param mixed
+     * @param mixed ...
      * @return \Mockery\Expectation
      */
-    public function shouldNotReceive()
+    public function shouldNotReceive($args)
     {
         $expectation = call_user_func_array(array($this, 'shouldReceive'), func_get_args());
         $expectation->never();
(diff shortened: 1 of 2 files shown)


aik099:
No parameters are listed in the interface, because it can accept any number of parameters.

padraic:
Looks fine and will see what @davedevelopment thinks.

GrahamCampbell:
This is a breaking change. We really need to get the branch alias sorted out.

davedevelopment:
:+1: from me

padraic:
It could be marked to 1.0.0 for that eventual bump, but I don't see the breakage as anything major. It's primarily a correction for documentation purposes and we are within the dev integers version wise. Is there any usage where these methods receive no parameters at all to make it breaking beyond that?

fgrosse:
@aik099 
> No parameters are listed in the interface, because it can accept any number of parameters.

Out of curiosity: what are the semantics of :

```php
$myMock->shouldReceive();
```

padraic:
@fgrosse It's $arg,... Technically, the doc line should note when the first is optional. I'm not sure it would be accurate though. The methods were intended to receive at least one argument.

fgrosse:
@padraic I had the argument name to `$arg` in my initial commit 520d08ca80ae8e7e5d554c984b8ba5a0b9c35d94 but I decided against it to make the interface clearer to anybody who might want to implement/use that.

After all passing only a single function is probably by far the biggest use case and the possibility of passing multiple method names there is documented in phpdoc.

fgrosse:
@padraic whats the status here? anything I can do to improve this PR? I would *really* like all the warnings in PhpStorm to disappear

daniphp:
I used the implementation \Mockery\Mock instead of \Mockery\MockInterface. I guess it scans the actual code and sees the var arguments.

fgrosse:
@daniphp I could confirm that this makes the warning go away on PhpStorm PS-141.1717
However I still think the interface should be updated. I can image I am not the only one who is annoyed by having warnings all over the test code because of this.

davedevelopment:
Their you go, will see what happens.

PR #471, opened 2015-05-12
Makes callback in ->with() be called only once per expectation

### LINKED ISSUE #468: Calling a degenerative `->with()`
Hello,
I have a test that checks that a method has been called with some specific arguments. Each argument must be passed only once.
Here's the failling test case:

```php
public function testBug() {
	$expectedZipNames = ['name1', 'name2', 'name3'];

	$zip = \Mockery::mock('zip', function($mock) use (&$expectedZipNames) {
		$mock->shouldReceive('getFromName')
			->times(2)
			->with(\Mockery::on(function($argument) use (&$expectedZipNames)  {
				// check if the argument is in the valid list, and remove it to not be able to validate it again.
				if (($key = array_search($argument, $expectedZipNames)) !== false) {
					array_splice($expectedZipNames, $key, 1);
					return true;
				}
				return false;
			}))
			->andReturn(false);
	});

	$zip->getFromName('name1');
	$zip->getFromName('name2');
	// The error show should be
	// Mockery\Exception\InvalidCountException:
	// 		Method getFromName(object(Mockery\Matcher\Closure)) from Mockery_0__zip
	// 		should be called exactly 2 times but called 3 times.
	// but I get
	// Mockery\Exception\NoMatchingExpectationException:
	// 		No matching handler found for Mockery_0__zip::getFromName("name3").
	// 		Either the method was unexpected or its arguments matched no expected argument list for this method
	$zip->getFromName('name3');

	$this->assertCount(0, $expectedZipNames);
}
```

I think that it can be fixed by changing ExpectationDirector@_findExpectationIn
https://github.com/padraic/mockery/blob/master/library/Mockery/ExpectationDirector.php#L173
`if ($exp->matchArgs($args) && $exp->isEligible()) {`
to
`if ($exp->isEligible() && $exp->matchArgs($args)) {`
but I do not know all the underlying implications.

Thanks.

vizath:
Fixes #468

### DIFF
--- a/library/Mockery/ExpectationDirector.php
+++ b/library/Mockery/ExpectationDirector.php
@@ -170,7 +170,7 @@ class ExpectationDirector
     protected function _findExpectationIn(array $expectations, array $args)
     {
         foreach ($expectations as $exp) {
-            if ($exp->matchArgs($args) && $exp->isEligible()) {
+            if ($exp->isEligible() && $exp->matchArgs($args)) {
                 return $exp;
             }
         }
--- a/tests/Mockery/ExpectationTest.php
+++ b/tests/Mockery/ExpectationTest.php
@@ -680,7 +680,7 @@ class ExpectationTest extends MockeryTestCase
         $number_of_calls = 0;
         $this->mock->shouldReceive('foo')
             ->times(2)
-            ->with(\Mockery::on(function($argument) use (&$number_of_calls) {
+            ->with(\Mockery::on(function ($argument) use (&$number_of_calls) {
                 $number_of_calls++;
                 return $number_of_calls <= 3;
             }));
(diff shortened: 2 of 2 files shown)


aik099:
Isn't `_findExpectationIn` method called only once anyway?

vizath:
Yes, but there is 2 calls to `matchArgs` in `_findExpectationIn`. If the Count expectation failed, the callback was called in both foreachs. By switching the order of the calls, when the count fails, the first `matchArgs` is not called.

I'm pretty sure there is a bigger underlying problem and a better way to solve this issue.
For example, if there is two calls to `shouldReceive`, the callback is still called the wrong number of  times.
```javascript
$number_of_calls = 0;

$this->mock->shouldReceive('foo')
    ->times(2);

$this->mock->shouldReceive('foo')
    ->with(\Mockery::on(function ($argument) use (&$number_of_calls) {
        $number_of_calls++;
        return $number_of_calls <= 3;
    }));

$this->mock->foo(1);
$this->mock->foo(1);
$this->mock->foo(1);
$this->container->mockery_verify();
```
With this example, there is _no exeption thrown_, even if the called count is higher than 2. And the callback is called _only once_.
I do not know if we should never call `shouldReceive` twice. The doc is unclear about this and the code seems to be able to manage it. My patch do not fix this perticular new issue.

Issue #355, opened 2014-07-12
Mock always calls __destruct (PHP 5.4/5.5, any Mockery version)


tiraeth:
Hi,

I've encountered an error with `__destructor` being called regardless constructor call setup. See code below:

``` php
<?php

require_once __DIR__ . '/vendor/autoload.php';

class Foo
{
    public function __construct()
    {
        echo 'Foo::__construct' . PHP_EOL;
    }

    public function __destruct()
    {
        echo 'Foo::__destruct' . PHP_EOL;
    }

    public function bar()
    {
    }
}

// ...
```

First attempt (incorrect behaviour):

``` php
<?php

// ...

$fooMock = Mockery::mock('Foo');

Mockery::close();
```

Expected output: **(empty)**
Received output: **Foo::__destruct\n**

Second attempt (correct behaviour):

``` php
<?php

// ...

$fooMock = Mockery::mock('Foo[bar]');

Mockery::close();
```

Expected output: **Foo::__construct\nFoo::__destruct\n**
Received output: **Foo::__construct\nFoo::__destruct\n**

Environment: PHP `5.5.14`, Mockery `dev-master` (and `0.9.1`), doesn't matter if I have XDebug enabled or not.

I've found that `__destruct` is blacklisted in `$builder` but still I'm curious why is it called. Mockery or PHP bug?

Appreciate your help, thanks!

**Edit:** Checked on OS X's bundled PHP `5.4.24` and got the same output.

**Edit 2:** Checked `0.9.0`, `0.8.0`, and `0.7.2` versions of mockery. Still the same result.

antonienko:
Any news on this issue?

robertbasic:
This was implemented by #215.

PR #479, opened 2015-06-24
#215 Original class destructor is called only when original construct…

davedevelopment:
…or was being called

This is a rebase of #218 and is related to #215 #355

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Change Log
 
+## 1.0.0 (XXXX-XX-XX)
+
+* Destructors (`__destruct`) are stubbed out where it makes sense
+
 ## 0.9.4 (XXXX-XX-XX)
 
 * `shouldIgnoreMissing` will respect global `allowMockingNonExistentMethods`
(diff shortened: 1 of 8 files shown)


davedevelopment:
#215, #218 and #355 are pretty old issues that I (and others) have failed to come to a consensus on, so I'm just going to go ahead and merge this if nobody comes up with some strong objections today.

davedevelopment:
I got impatient

GrahamCampbell:
lol :)

reinink:
I just ran into this issue when trying to mock the Mandrill API client library. I see this fix is tagged for version 1. Do you know of any work around to the problem in the meantime?

davedevelopment:
@reinink this should get around it, but may have some side effects and probably isn't considered public API.

``` php
$mock = Mockery::mock(new \Mockery\Generator\MockConfiguration([MyThing::class]));
```

reinink:
@davedevelopment Hey thanks, but I've actually just used a recent dev-master version, which seems to be working well.

sanderdg:
@davedevelopment I'm also ran into this issue today and your suggested fix doesn't seem to work for me:
`
$mandrill = Mockery::mock(new MockConfiguration([Mandrill::class]));
die(($mandrill instanceof Mandrill) ? 'It works' : 'It doesn\'t work');
`

erikverheij:
@sanderdg, I ran in the same issue, also with setting up a test for Mandrill eheh 😄.

I managed to work around it by setting a fake curl handle which makes the original destructor work again.
```
$mandrillClientMock->ch = curl_init('http://www.some-dummy-url.com/');
```

PR #482, opened 2015-06-26
allow mocks to be overwritten

maxgalbu:
Related to https://github.com/padraic/mockery/issues/309

### DIFF
--- a/library/Mockery/Container.php
+++ b/library/Mockery/Container.php
@@ -71,7 +71,7 @@ class Container
      *
      * @var boolean
      */
-    public static $_overwriteMocks = true;
+    public static $_overwriteMocks = false;
 
     /**
      * @var array
--- a/library/Mockery/Container.php
+++ b/library/Mockery/Container.php
@@ -67,6 +67,13 @@ class Container
     protected $_loader;
 
     /**
+     * Should we overwrite mocks of the same class
+     * 
+     * @var boolean
+     */
+    public static $_overwriteMocks = true;
+
+    /**
      * @var array
      */
     protected $_namedMocks = array();
@@ -75,6 +82,16 @@ class Container
     {
         $this->_generator = $generator ?: \Mockery::getDefaultGenerator();
         $this->_loader = $loader ?: \Mockery::getDefaultLoader();
+    }
+
+    /**
+     * Set wether the mocks of the same class should overwrite the previos
+     * defined mocks
+     * 
+     * @param boolean $overwriteMocks
+     */
+    public static function setOverwriteMocks($overwriteMocks) {
+      self::$_overwriteMocks = $overwriteMocks;
     }
 
     /**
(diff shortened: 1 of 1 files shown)


aik099:
Doesn't sound right because it would leave "allow overwrite" flag and next expectations will replace existing ones all the time if flag isn't removed.

If it's a must have feature (@davedevelopment, what you think of this feature?) then instead I would introduce a method, that would clear previously added expectations when called. Not sure what to do with data recorded (in case of spies), when expectation to be deleted were already matched. We might have some issues with "assertion count" number at the end of the test.

library/Mockery/Container.php
```diff
@@ -67,6 +67,13 @@ class Container
     protected $_loader;
 
     /**
+     * Should we overwrite mocks of the same class?
+     *
+     * @var boolean
```
GrahamCampbell: bool not boolean

maxgalbu:
@aik099 yeah sorry should be false by default

aik099:
@maxgalbu , totally missed that `false` thing. But anyway I belive clearing expectations once is better then `enable flag` + `set 1 expectation` + `disable flag`.

davedevelopment:
Could you possibly explain the use case for this?

maxgalbu:
I'm using mockery to mock classes used by other classes, so I don't actually use expectations after mocking. I have a test for a class that checks for user permissions, and I need to change (in the same test) the class that returns those user permission to force a specific permission. Without this change mockery simply adds subsequent mocks in the $this->_mocks array as indexes:

https://github.com/padraic/mockery/blob/3aee65d0a6a9b5e3e8b95a4a4b7830257e7ed608/library/Mockery/Container.php#L419-L423

davedevelopment:
Sorry, I still don't fully understand the use case, could you post some code? Would something like this work?

``` php
$mock = Mockery::mock("overload:SomeClass");

$mock->shouldReceive("getPermissions")->andReturn(644)->byDefault();
var_dump((new SomeClass)->getPermissions()); // 644
$mock->shouldReceive("getPermissions")->andReturn(666)->byDefault();
var_dump((new SomeClass)->getPermissions()); // 666
$mock->shouldReceive("getPermissions")->andReturn(640)->byDefault();
var_dump((new SomeClass)->getPermissions()); // 640
```

See #483

maxgalbu:
It's working for a simple script, but not in the actual test. The only different thing from my test is `byDefault()`

davedevelopment:
I don't understand what you mean, are you saying that is the kind of thing you're looking for?

maxgalbu:
Yes, I probably already tried with your code, but since it wasn't working on my script i looked for another way :D so i made this pull request

davedevelopment:
Cool, so I just merged #483, which should make the `byDefault` work for instance mocks, you'll just need to depend on master until a release gets put together

maxgalbu:
Actually the byDefault was working even without using dev-master from composer, I don't really know what's going on in my tests

Squeezya:
i use codeception for Laravel tests and i use mockery, what i do is 
```
if (self::$myMock) {
    Mockery::close();
}   
self::$myMock = Mockery::mock('alias:myMockClass'); ->with namespace
self::$myMock->shouldReceive('user')->andReturn(new User([]))->byDefault();
```
the close allow to clear the mocks

PR #483, opened 2015-06-26
Instance mocks should receive a copy of default expectations

davedevelopment:


davedevelopment:
Hmm, might need some more work on this

aik099:
What are these `default expectations` and what is the issue solved by this PR actually?

davedevelopment:
Default expectations are ones that can be overridden, so they're almost secondary expectations. When we check to see if we were expecting something, we loop over the normal expectations first, then over the default expectations.

aik099:
Ah, I thought that default expectations are ones, that are added first and when you add `->byDefault()` the expectation is added to the top of the expectation list instead of to the end as happens usually.

PR #486, opened 2015-07-06
Add note about shouldDeferMissing

jatubio:
I think it's good to clarify that real parent's method's will be called if expected parameters signature not match.


davedevelopment:
Really sorry I left this open for so long, I'll be honest and say I don't remember seeing it come in...

I'm not sure it clarifies it all that well, it took me a couple of reads to understand.

Issue #487, opened 2015-07-09
My /tmp is filled with mockery files, what am I missing?


marcaube:
It seems a *temporary* file is generated (e.g. `/tmp/MockeryvAahXH`) in [ClassPass](https://github.com/padraic/mockery/blob/master/library/Mockery/Generator/StringManipulation/Pass/ClassPass.php#L36) for each class mocks. The problem I have is that they are never deleted.  There's tens of thousands of them.

My test classes extend `\Mockery\Adapter\Phpunit\MockeryTestCase`, am I missing something? Do I still need the `tearDown()` method or the listener?

davedevelopment:
I think creating a file is a necessary evil there, but ideally it's a narrow use case, that should only happen whenever the class you are trying to mock doesn't actually exist or at least can't be found by autoloading. 

I'd be happy to merge something that attempts a clean up on shutdown or something.

marcaube:
Ah! well, that is exactly my case. Some of the classes that are mocked do not exist in the test context. They are autoloaded deep inside a legacy system, and the unit tests steer clear of that bootstrap.

If I were to code something for this, how should I tackle the problem? Keep track of all the temporary files in an array somewhere and unlink them during shutdown?

davedevelopment:
> If I were to code something for this, how should I tackle the problem? Keep track of all the temporary files in an array somewhere and unlink them during shutdown?

Yeah I think that would do it. You could hook it in to `\Mockery::teardown`, that will probably be cleaner than trying to use php's `register_shutdown_function`.

``` php 
// in ClassPass
\Mockery::registerFileForCleanUp($tmpfname);

// in Mockery::tearDown
foreach (static::$filesForCleanUp as $filename) {
    @unlink($filename);
}
```

marcaube:
Thanks, I'll try to come up with a tested solution before the end of the week. I love That Podcast BTW.

GrahamCampbell:
> 
// in Mockery::tearDown

Why not just have mockery do this in it's destructor?

marcaube:
You mean in `Mock::__destruct()`?

aik099:
> I love That Podcast BTW.

What podcast?

davedevelopment:
@GrahamCampbell I'd prefer the explicit over implicit.

@aik099 *That Podcast* (thatpodcast.io)

GrahamCampbell:
> You mean in Mock::__destruct()?

No. PHP calls that for you whenever an object is being deleted.

[...]

PR #488, opened 2015-07-10
Clean-up temporary files on teardown

marcaube:
When you mock a class that does not exist, or can't be found by autoloading, Mockery creates a temporary file, so the class can be required. The clean-up was left to the OS.

This change registers the files to be cleaned-up on teardown by Mockery.

See issue #487.

### DIFF
--- a/library/Mockery.php
+++ b/library/Mockery.php
@@ -64,6 +64,11 @@ class Mockery
     protected static $_loader;
 
     /**
+     * @var array
+     */
+    private static $_filesToCleanUp = [];
+
+    /**
      * Static shortcut to \Mockery\Container::mock().
      *
      * @return \Mockery\MockInterface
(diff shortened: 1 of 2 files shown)


davedevelopment:
Thanks! Great job

ricardofandrade:
@davedevelopment would it be possible to backport this to the 0.9 branch?

davedevelopment:
@ricardofandrade I haven't got a lot of time right now, but if someone could give it a try and test, I'll hit the merge button

ricardofandrade:
@davedevelopment I appreciate you reply!
I've patched my environment with the changes and they work great.
If you prefer a more formal approach, I can fork, cherry-pick and submit a PR against 0.9.

davedevelopment:
@ricardofandrade if you could, that would be great! I can just hit the merge button that way :)

PR #489, opened 2015-07-12
Add support for parameter and return types

trowski:
This PR adds support for scalar parameter types return types on mocked object methods while maintaining compatibility with older PHP versions.

If a return type is declared on a method and no return expectation has been provided for the mock method call, a return value is automatically generated based on the declared type. This includes creating a new mock object to return if necessary.

### DIFF
--- a/library/Mockery/Expectation.php
+++ b/library/Mockery/Expectation.php
@@ -216,6 +216,32 @@ class Expectation implements ExpectationInterface
         } elseif (count($this->_returnQueue) > 0) {
             return current($this->_returnQueue);
         }
+
+        $rm = $this->_mock->mockery_getMethod($this->_name);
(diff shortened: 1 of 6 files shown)


davedevelopment:
This is awesome, thank you

library/Mockery/Expectation.php
```diff
@@ -216,6 +216,32 @@ protected function _getReturnValue(array $args)
...
+                case 'int':    return 0;
+                case 'float':  return 0.0;
+                case 'bool':   return false;
+                case 'array':  return [];
+
+                case 'callable':
+                case 'Closure':
+                    return function () {};
+
+                case 'Traversable':
```
aik099: Iterators are also traversable and we can support them separately when PHP 5.4 is used.

trowski: Iterators are handled by the call below to `Mockery::mock()` (and so can `Traversable` as it turns out, but it's probably faster to just return a generator).

Issue #490, opened 2015-07-13
ShouldIgnoreMissing/Spies should respect return type hints


davedevelopment:
`shouldIgnoreMissing` (and therefore spies) should attempt to respect return type hints, to avoid type errors and make mockery more enjoyable.

```php
class Dave
{
    public function foo(): string {}
}

$mock = m::mock(Dave::class)->shouldIgnoreMissing();

var_dump($mock->foo()); // string(0) ''

```

See #489

aik099:
Isn't a `null`, that is currently returned, pass all type hints already?

davedevelopment:
I'm not sure what happened to the nullable types RFC, but as far as I know you'll need to explicitly say if null is allowed, so we'll check for that too when the time comes.

trowski:
@aik099 Returning `null` is not acceptable if a return type has been declared, it results in a `TypeError` being thrown. The nullable types RFC is a possibility for 7.1. If it happens, I'm sure a method like `isReturnNullable()` will be added to `ReflectionMethod`. If that happens, `null` could be returned for those methods.

@davedevelopment You could use `null` to indicate that the return type should be automatically generated. This seems to make sense and would make existing tests work automatically if someone adds a return type to a method.

Changing [`library/Mockery/Mock.php` at line 710](https://github.com/padraic/mockery/blob/master/library/Mockery/Mock.php#L710) to the following makes the code you post above result in the expected `string(0) ""`.

```php
} else if (null === $this->_mockery_defaultReturnValue) {
    $rm = $this->mockery_getMethod($method);
    if ($rm && version_compare(PHP_VERSION, '7.0.0-dev') >= 0 && $rm->hasReturnType()) {
        $type = (string) $rm->getReturnType();
        switch ($type) {
            case '':       return;
            case 'string': return '';
            case 'int':    return 0;
            case 'float':  return 0.0;
            case 'bool':   return false;
            case 'array':  return [];

            case 'callable':
            case 'Closure':
                return function () {};

            case 'Traversable':
            case 'Generator':
                // Remove eval() when minimum version >=5.5
                $generator = eval('return function () { yield; };');
                return $generator();

            default:
                return \Mockery::mock($type);
        }
    }
    return;
} else {
    return $this->_mockery_defaultReturnValue;
}
```

This results in a bit of code duplication. I could consolidate the type generation somewhere if you have a suggestion and do another PR.

davedevelopment:
[...]

PR #492, opened 2015-07-23
Mock return types on spies

trowski:
This PR addresses #490 by creating a mock return value when `shouldIgnoreMissing()` is called on the mock object and subsequently a method is called on the mock without setting expectations.


davedevelopment:
Awesome, thank you!

PR #495, opened 2015-07-31
Try and throw if method names aren't valid

### LINKED ISSUE #494: Blank should receive
My test had `$mock->shouldReceive()->once()` instead of `$mock->shouldReceive('method')->once()` which because my with was quite large I didn't notice. Maybe we should throw an exception if no method name is passed?

image

Recommended by @davedevelopment

davedevelopment:
Fixes #494

### DIFF
--- a/library/Mockery/Mock.php
+++ b/library/Mockery/Mock.php
@@ -172,6 +172,16 @@ class Mock implements MockInterface
      */
     public function shouldReceive($methodName)
     {
+        if (func_num_args() < 1) {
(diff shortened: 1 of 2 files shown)


bweston92:
Maybe just check the method names are not empty and not the arguments length, as php will throw it's own error. The issue in #494 is that no versions are tagged with the latest `shouldReceive` signature but now it requires 1 argument minimum then php itself will of course raise an error.

PR #501, opened 2015-08-25
Add AnyTypeOf matcher

### LINKED ISSUE #500: Add AnyTypeOf matcher
With the current codebase, I can check an individual argument either against a specific type, or against a set of concrete values:

```php
$some_object = Mockery::mock( 'SomeClass' );
$some_object->shouldReceive( 'foo' )->with( Mockery::type( 'int' ) );
$some_object->foo( 1 );

$some_object->shouldReceive( 'foo' )->with( Mockery::anyOf( 0, 1, 2 ) );
$some_object->foo( 1 );
```

What's missing, however, is *checking against a set of types* so one could do something like the following:

```php
$some_object = Mockery::mock( 'SomeClass' );
$some_object->shouldReceive( 'foo' )->with( Mockery::anyTypeOf( 'int', 'string' ) );
$some_object->foo( 1 );
```

I'll send a pull request for this in a minute.

tfrommen:
This pull request fixes #500.

What this pull request does:
* allow for *checking against a set of types*.

```php
$some_object = Mockery::mock( 'SomeClass' );
$some_object->shouldReceive( 'foo' )->with( Mockery::anyTypeOf( 'int', 'string' ) );
$some_object->foo( 1 );
```

I also included tests for the new matcher, however, due to the nature of the matcher you could come up with infinite test cases...

BTW, one could also add the according counterpart `notAnyTypeOf`.

### DIFF
--- a/library/Mockery.php
+++ b/library/Mockery.php
@@ -395,6 +395,16 @@ class Mockery
     }
 
     /**
+     * Return instance of ANYTYPEOF matcher.
+     *
+     * @return \Mockery\Matcher\AnyTypeOf
+     */
+    public static function anyTypeOf()
+    {
+        return new \Mockery\Matcher\AnyTypeOf(func_get_args());
+    }
+
+    /**
      * Get the global configuration container.
      */
     public static function getConfiguration()
--- a/library/Mockery/Matcher/AnyTypeOf.php
+++ b/library/Mockery/Matcher/AnyTypeOf.php
@@ -1,0 +1,51 @@
+<?php
+/**
+ * Mockery
+ *
+ * LICENSE
+ *
+ * This source file is subject to the new BSD license that is bundled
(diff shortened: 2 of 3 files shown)


aik099:
Thanks, but I'd recommend using the bundled matchers from Hamcrest library instead of duplicating it's functionality within Mockery.

tfrommen:
Hm, okay, that's good (and at the same time too bad) that it's already possible. I'm not using Mockery as part of the WordPress-specific mocking tool WP_Mock, that's why I'm not that familiar with Mockery's dependencies and such things. So sorry for the PR. :|
I'll try to find how I can do what I want to with Hamcrest then.

aik099:
We still have matchers in mockery for BC reasons. Maybe we should internally make these matchers to use Hamcrest functions :smile:

PR #504, opened 2015-09-10
Manually build up argument list

### LINKED ISSUE #503: NoMatchingExpectationException removes spaces, causes frustration.
A lot.

I've spent some time figuring out why this code didn't work;

```
$this->mockedObject->shouldReceive('push')->with('foobar');
```

```
1) Test\MyTest
Mockery\Exception\NoMatchingExpectationException: No matching handler found for Mockery_7_Test::push("foobar"). Either the method was unexpected or its arguments matched no expected argument list for this method
```

HOWEVER it turned out that my code was not sending "foobar" but "foo bar" and the spaces were stripped from the output. 

```
$this->queue->shouldReceive('push')->with('foo bar');
```

```
OK (1 test, 3 assertions)
```

I understand where this might come from (because huge dumps are hard to read) but this is very frustrating. 

I would suggest not stripping spaces - or any whitespace at all, to prevent this kind of stuff.

davedevelopment:
Fixes #503


library/Mockery.php
```diff
@@ -440,16 +440,19 @@ private static function formatArgument($argument, $depth = 0)
...
             if ($depth === 1) {
-                $argument = 'array(...)';
+                $argument = '[...]';
             } else {
                 $sample = array();
                 foreach ($argument as $key => $value) {
-                    $sample[$key] = self::formatArgument($value, $depth + 1);
+                    $key = is_int($key) ? $key : "'$key'";
+                    $value = self::formatArgument($value, $depth + 1);
+                    $sample[]= "$key => $value";
```
GrahamCampbell: cs

davedevelopment: Ta G

GrahamCampbell:
This needs rebasing to load in the previous fixes to get stuff passing.

tests/Mockery/ContainerTest.php
```diff
@@ -1270,6 +1270,22 @@ public function testHandlesMethodWithArgumentExpectationWhenCalledWithNestedReso
     }
 
     /**
+     */
```
GrahamCampbell: empty phpdoc?

tests/Mockery/ContainerTest.php
```diff
@@ -1270,6 +1270,22 @@ public function testHandlesMethodWithArgumentExpectationWhenCalledWithNestedReso
...
+        $mock->shouldReceive("foo")->with(123);
+
+        $this->setExpectedException(
+            "Mockery\Exception\NoMatchingExpectationException",
+            "MyTestClass::foo(true, false, [0 => true, 1 => false])"
+        );
+
+        $mock->foo(true, false, [true, false]);
+    }
+
```
GrahamCampbell: too many newlines?

RobQuistNL:
Test case:
```php
<?php

$mock = Mockery::mock();
$mock->shouldReceive('push')->with('value', ['var2' => 'true']);

$mock->push('value', ['var2' => true, 'var 123' => 'with spaces']);
```

Master:
```
PHP Fatal error:  Uncaught exception 'Mockery\Exception\NoMatchingExpectationException' with message 'No matching handler found for Mockery_0::push("value", array('var2'=>'true','var123'=>'withspaces',)). Either the method was unexpected or its arguments matched no expected argument list for this method
```

This PR
```
PHP Fatal error:  Uncaught exception 'Mockery\Exception\NoMatchingExpectationException' with message 'No matching handler found for Mockery_0::push('value', ['var2' => true, 'var 123' => 'with spaces']). Either the method was unexpected or its arguments matched no expected argument list for this method
```

:+1: from me! Thanks for putting your time and effort into this.

PR #510, opened 2015-10-02
Added missing quotes to a parameter

gammamatrix:
The method name is passed as a string: `$mock->shouldReceive('method_name')`


davedevelopment:
Thanks!

PR #513, opened 2015-10-10
Prevent trigger of error when exception occurs during unserialisation

### LINKED ISSUE #512: Fatal error is triggered when error or exception occurs during instantiation using unserialisation
A fatal error is triggered when an exception occurs during instantiation using unserialisation:

> PHP Fatal error:  Cannot use Exception as Exception because the name is already in use

Runtime:	PHP 5.6.14-1+deb.sury.org~trusty+1 with Xdebug 2.3.2

The cause of this error is that `\Exception` is imported as `Exception` in `\Mockery\Instantiator`, while a class name `Exception` already exists in that namespace (`\Mockery\Exception`). The solution appears to be to apply the below patch:

```diff
diff --git a/library/Mockery/Instantiator.php b/library/Mockery/Instantiator.php
index 9006496..f396fd8 100644
--- a/library/Mockery/Instantiator.php
+++ b/library/Mockery/Instantiator.php
@@ -20,7 +20,6 @@
 namespace Mockery;
 
 use Closure;
-use Exception;
 use ReflectionClass;
 use UnexpectedValueException;
 use InvalidArgumentException;
@@ -135,7 +134,7 @@ final class Instantiator
 
         try {
             unserialize($serializedString);
-        } catch (Exception $exception) {
+        } catch (\Exception $exception) {
             restore_error_handler();
 
             throw new UnexpectedValueException("An exception was raised while trying to instantiate an instance of \"{$reflectionClass->getName()}\" via un-serialization", 0, $exception);

```

A fully qualified reference to `\Exception` is already used five lines above the catch clause.

Would you like me to open a PR with this patch? I can provide a regression test if so required.

rjkip:
Fixes #512:

A fatal error is triggered when an exception occurs during instantiation using unserialisation:

> PHP Fatal error: Cannot use Exception as Exception because the name is already in use
Runtime: PHP 5.6.14-1+deb.sury.org~trusty+1 with Xdebug 2.3.2

The cause of this error is that \Exception is imported as Exception in \Mockery\Instantiator, while a class name Exception already exists in that namespace (\Mockery\Exception).

This commit fixes the name collision. I don't quite remember how I got a error to trigger during unserialisation, so there's no regression test.

### DIFF
--- a/library/Mockery/Instantiator.php
+++ b/library/Mockery/Instantiator.php
@@ -20,7 +20,6 @@
 namespace Mockery;
 
 use Closure;
-use Exception;
 use ReflectionClass;
 use UnexpectedValueException;
 use InvalidArgumentException;
@@ -135,7 +134,7 @@ final class Instantiator
 
         try {
             unserialize($serializedString);
-        } catch (Exception $exception) {
+        } catch (\Exception $exception) {
             restore_error_handler();
 
             throw new UnexpectedValueException("An exception was raised while trying to instantiate an instance of \"{$reflectionClass->getName()}\" via un-serialization", 0, $exception);


davedevelopment:
Thanks!

athlan:
Still appears in version 0.9.5.

```
$ phpunit
PHPUnit 5.4.6 by Sebastian Bergmann and contributors.

EEEE
Fatal error: Cannot use Exception as Exception because the name is already in use in E:\ ... \vendor\mockery\mockery\library\Mockery\Instantiator.php on line 23
```

PR #514, opened 2015-10-11
Update README.md

Jean85:
Update dates and version support. The closing of the PEAR channel is now in the past, and the master branch should be proposed, not optional.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,8 +17,8 @@ phpunit-mock-objects without the World ending.
 Mockery is released under a New BSD License.
 
 The current released version on Packagist is `0.9.4`.
-The current released version for PEAR is `0.9.0`. Composer users may instead opt to use
-the current master branch aliased to `1.0.x-dev`.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -17,8 +17,8 @@ phpunit-mock-objects without the World ending.
 Mockery is released under a New BSD License.
 
 The current released version on Packagist is `0.9.4`.
-The current released version for PEAR is `0.9.0`. Composer users may instead opt to use
-the current master branch aliased to `1.0.x-dev`.
+Composer users may opt to use the current master branch aliased to `1.0.x-dev`.
```
GrahamCampbell: No? They either want `0.9.*` or `~1.0`. No need for any `dev-master` or anything else.

Jean85: ~1.0 with minimum stability stable will fail, then...

GrahamCampbell: `~1.0@dev` then

Jean85:
Fixed as suggested by @GrahamCampbell

PR #521, opened 2015-11-17
Improved efficiency of HasKey matcher.

Bilge:


davedevelopment:
I think this would have undesired effects when a key has a value of null?

aik099:
You can use `array_key_exists` as well. In the past `isset` proved to be faster then `array_key_exists`, but performance tests done today doesn't confirm that anymore.

Bilge:
@davedevelopment How so?

davedevelopment:
``` php
$arr = [
    'some_key' => null,
];

var_dump(isset($arr['some_key'])); // bool(false)
```

Bilge:
@davedevelopment Oops! Then I think we should use @aik099's suggestion.

davedevelopment:
Thanks!

PR #524, opened 2015-12-11
Add to "require-dev" instead of "require" in composer.json

vlakoff:
Also for consistency with [Installation guide](https://github.com/padraic/mockery/blob/master/docs/getting_started/installation.rst).


davedevelopment:
Thanks

PR #526, opened 2015-12-14
fix #472 throw exception if given class name is invalid

### LINKED ISSUE #472: Readable error for invalid class name passed to Mockery::mock()
Assuming a code like:

```php
Mockery::mock( 'foo.class' );
```

is perfectly valid PHP code, but because `foo.class` does not exists (it can't exist) Mockery just create a temporary file that contains something like:

```php
class foo.class {}
```

This, of course, result in parse error because a class name can't contain a dot.

```
Parse error: syntax error, unexpected '.',
expecting '{' in /path/to/tmp/moc_dfes.tmp on line 3
``` 

A similar error occurs for any invalid class names passed to `Mockery::mock`.

I think Mockery should throw an exception with a more easily understandable message when an invalid class name is passed to `Mockery::mock`, otherwise to actually understand why tests fail is to open the temporary file, read it, and try to figure out where and why it was created (and if the invalid name is because of a typo it is everything but simple).

nobuf:
This PR is attempting to fix #472. While Mockery gets a parse error with `Mockery::mock('this-is-invalid')` this fix throws a Mockery\Exception error with friendlier message. It uses [the regex on the official document](http://php.net/manual/en/language.oop5.basic.php) and I'm assuming that namespaces follow the same rule.

It might be a little too strict since it currently allows a parameter like `mock('\\\\Foo')`. Please let me know if there's any better way.

### DIFF
--- a/library/Mockery/Container.php
+++ b/library/Mockery/Container.php
@@ -539,8 +539,9 @@ class Container
             $className = substr($className, 1); // remove the first backslash
         }
         // all the namespaces and class name should match the regex
-        return empty(array_filter(explode('\\', $className), function ($name) {
+        $invalidNames = array_filter(explode('\\', $className), function ($name) {
             return !preg_match('/^[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*$/', $name);
-        }));
+        });
+        return empty($invalidNames);
     }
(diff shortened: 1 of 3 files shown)


davedevelopment:
Looks great, thank you!

PR #527, opened 2015-12-31
Compare classes with deprecated methods doesn't work

Myiyk:
Mockery try to compare two objects (same type), but some methods are deprecated. And calling deprecated methods make errors.

In my opinion deprecated methods need to be skipped in comparing.

This PR is only for show example, where is problem.


tests/Mockery/AdhocTest.php
```diff
@@ -82,16 +82,25 @@ public function testMockeryConstructAndDestructIsNotCalled()
         $this->container->mockery_close();
         $this->assertFalse(MockeryTest_NameOfExistingClassWithDestructor::$isDestructorWasCalled);
     }
-
```
GrahamCampbell: ?

robertbasic:
Ever since #732 got merged, this shouldn't be a problem any more as we don't do random `get*` and `is*` calls.

Closing this PR.

Issue #464, opened 2015-04-07
Hamcrest conflicts with PHPUnit


Ocramius:
As it stands, hamcrest-php conflicts with PHPUnit:
- [`any` defined by hamcrest](https://github.com/hamcrest/hamcrest-php/blob/18ec485a0696da23a0d5d2b21b1961d77e42d27d/hamcrest/Hamcrest.php#L362)
- [`any` defined by phpunit](https://github.com/sebastianbergmann/phpunit/blob/8044ba2d89cb7c9c5ba969403e53191eb7eb9aca/src/Framework/Assert/Functions.php#L18)

Both nicely _NOT_ namespaced

_insert long sigh here_

aik099:
I wonder how Mockery unit tests pass knowing that they use Hamcrest and PHPUnit together.

GrahamCampbell:
Not a conflict: https://github.com/sebastianbergmann/phpunit/blob/8044ba2d89cb7c9c5ba969403e53191eb7eb9aca/src/Framework/Assert/Functions.php#L18

GrahamCampbell:
hamcrest defines them only if phpunit has't defined that function

aik099:
But if libraries are included in different order: first Hamcrest and then PHPUnit, then all fails.

Ocramius:
Depending on autoloader priority, there may or may not be a crash. I had a crash this morning, heh...

aik099:
If Hamcrest conflicts with PHPUnit, then shouldn't this be reported to Hamcrest repo? And initially we (at Hamcrest) stopped including global function definitions, like `any` and such in Composer (because we have namespaced class for that), but people started complaining that their usages of these functions no longer work. Then we added it to Composer auto-loader and that caused the problem.

GrahamCampbell:
phpunit fixed this in 4.6.2 i think?

Ocramius:
I mainly reported it here because the dependency was introduced in the latest minor version, as far as I know.

It can indeed be reported on Hamcrest and/or PHPUnit as well.

Possibly related: sebastianbergmann/phpunit#1625

aik099:
https://github.com/sebastianbergmann/phpunit/blob/master/src/Framework/Assert/Functions.php - not fixed. Same in 4.6 and 4.7 branches of PHPUnit.

padraic:
[...]

PR #535, opened 2016-01-19
Bump hamcrest/hamcrest to 2.0-dev

davedevelopment:
Refs #464

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
     "require": {
         "php": ">=5.4.0",
         "lib-pcre": ">=7.0",
-        "hamcrest/hamcrest-php": "~2.0-dev"
+        "hamcrest/hamcrest-php": "^2.0@dev"
     },
     "require-dev": {
         "phpunit/phpunit": "~4.0"
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
     "require": {
         "php": ">=5.4.0",
         "lib-pcre": ">=7.0",
-        "hamcrest/hamcrest-php": "~1.1"
+        "hamcrest/hamcrest-php": "~2.0-dev"
     },
     "require-dev": {
         "phpunit/phpunit": "~4.0"


composer.json
```diff
@@ -19,7 +19,7 @@
     "require": {
         "php": ">=5.4.0",
         "lib-pcre": ">=7.0",
-        "hamcrest/hamcrest-php": "~1.1"
+        "hamcrest/hamcrest-php": "~2.0-dev"
```
aik099: The constraint should say `^2.0@dev` instead. This way:

* you will be checking for a regular version
* for Mockery contributors the dev version of Hamcrest will be used
* for everybody else (who include Mockery as a dependency) stable 2.x version of Hamcrest will be used

davedevelopment: done

aik099:
Tests pass, so I guess it does solve the problem.

Or maybe we have some code left that orders auto-loader of PHPUnit and Hamcrest to play along? If that is the case, then we need to remove that and use new method, that autoloads Hamcrest global functions where needed in test suite of Mockery.

davedevelopment:
I don't think mockery's test suite was affected, as it only comes in to play if people manually attempt to load phpunit's assert functions.

warrickhill:
Any chance this could be reverted. Getting this when running composer update:
```
Problem 1
    - mockery/mockery dev-master requires hamcrest/hamcrest-php ^2.0@dev -> no matching package found.
```

davedevelopment:
@warrickhill what version of composer are you using? Do you have a custom repository set up for hamcrest?

warrickhill:
I have just run
```composer selfupdate
You are running composer with xdebug enabled. This has a major impact on runtime performance. See https://getcomposer.org/xdebug
Updating to version 593b88e41487f481d4368a4d331146914c1c2981.
    Downloading: 100%         
Use composer self-update --rollback to return to version 03299ff075236be27be356498d6c64def973fe41```

I am getting the problem on both versions. composer install also doesn't work.

No i'm not using a custom repo for hamcrest

chrismou:
@warrickhill Do you have "minimum-stability": "dev" set?  It won't pull through if you're set to stable only

davedevelopment:
Ah yeah, that would be the problem. I think you'll need to adjust the `minimum-stability` setting, or manually require `hamcrest/hamcrest-php: "^2.0@dev"` for now

warrickhill:
both those solutions have worked. I think we will have to manually require hamcrest for now. thanks

GrahamCampbell:
min stability dev, then prefer stable true is often a good combination

GrahamCampbell:
https://github.com/CachetHQ/Cachet/blob/v2.0.4/composer.json#L102-L103

Issue #539, opened 2016-01-27
andThrows with custom parameters


danhanly:
I have an exception that extends another, and assembles a message based upon, passing it to the parent constructor:

```php
class CustomException extends Exception
{
    public function __construct(array $response)
    {
        $message = "";
        foreach($response as $error) {
            $message = "Error: $error\n";
        }
        parent::__construct($message);
    }
}
```

This type of setup isn't able to be triggered via Mockery since the `andThrow` method forces a string message, not any other method.

Is there a way around this?

**What I've tried**

```php
$mock->andThrow(new CustomException(['test', 'test2']));
$mock->andThrow(CustomException::class, ['test', 'test2]);
```

davedevelopment:
Maybe you could new up the exception manually? I think that works.

``` php
$mock->shouldReceive("foo")->andThrow(new CustomException(['some error']));
```

danhanly:
Doesn't the `andThrow` method only accept a string as reference to the exception to be thrown?

I'll give it a spin.

danhanly:
This worked. Thanks

PR #540, opened 2016-01-27
andThrow Parameter Clarification

danhanly:
Response to #539 where intention wasn't clear. Just thought I'd clear up the PHPDoc to guide others not to make the same mistakes as I.


davedevelopment:
Thank you very much!

PR #542, opened 2016-01-28
Removed "--dev" option from composer install command

beni0888:
The "--dev" option of the `composer install` command is deprecated and it shows a warning message when you use it. In this way, I've removed every occurrence of this options that I've found in the project:

- Contribution guidelines
- Exception messages
- Code comments
- Scrutinizer config file


davedevelopment:
Thanks a bunch!

beni0888:
You're welcome ;)

PR #543, opened 2016-02-01
added constraint matcher

yjv:
adds a matcher you can pass any phpunit constrain to. makes it alot easier to use any phpunit constraint in arg matching

### DIFF
--- a/library/Mockery/Matcher/Constraint.php
+++ b/library/Mockery/Matcher/Constraint.php
@@ -1,7 +1,7 @@
 <?php
 namespace Mockery\Matcher;
 
-class ConstraintMatcher extends MatcherAbstract
+class Constraint extends MatcherAbstract
 {
     protected $constraint;
     protected $rethrow;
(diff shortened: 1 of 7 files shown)


aik099:
Why not use Hamcrest matchers instead?

yjv:
this will make it so in projects that arent using hamcrest also have this option since they will almost definitely have phpunit

davedevelopment:
Happy to have this in! Two things:

If we're going to be creating a separate class for it, I think it needs to be explicitly named `PHPUnitConstraint` or something. Alternatively, this could be done with a couple of lines in `Mockery\Expectation::_matchArg`, much like we do for Hamcrest matchers.

Needs an integration test or two as well, especially given the exception throwing nature of the matchers.

aik099:
Ah, so it's more like a wrapper to other library matchers, then a matcher of it's own?

Maybe in that case you should be changing code, where we're integrating with Hamcrest instead: https://github.com/padraic/mockery/blob/cc22d9bcd9fe9b34fc73587ed9bf2895b228138b/library/Mockery/Expectation.php#L334 ?

yjv:
@davedevelopment ok ill rename it. @aik099 i think i prefer implementing it as a matcher and leaving the expectation director more generic.

thanks!

davedevelopment:
Thanks!

Issue #459, opened 2015-03-23
Add `withArgs()` to validate via closure passed arguments at once


gmazzap:
I think would be handy a method like

``` php
withArgs($closure)
```

that would work similar to 

``` php
with( Mockery::on($closure) );
```

with the only difference that in first case the `$closure`receives **all** the arguments passed to the target method.

aik099:
With proposed `withArgs($closure)` syntax it's impossible to uniquely detect which expectation should fire because one with close will be used in all cases.

Instead I recommend using `->andReturnUsing($closure)`. Then the `$closure` receives all the arguments passed.

Giuseppe-Mazzapica:
Thanks @aik099 for your response.

To use `andReturnUsing` was my first thought, but to make test fail if arguments are not valid, the only way is to manually throw an exception, and it's quite verbose because an explicative message should be added too or is hard to understand why the test failed.

The only aim of my proposal is to have less verbose test code, so...

Well, yes, I assume that `withArgs` should be used as unique or last expectation for a method.

Thats IMO should not be a problem: for example, currently you can't mix `andReturnUsing` and `andReturn` in same test for same method, and in the same way you couldn't mix `with` and `withArgs`...

beni0888:
@padraic @aik099 @Giuseppe-Mazzapica So, what do you think about this feature? Should it be added to the API? Can I start working on it?

davedevelopment:
Can for me!

aik099:
How we can distinguish between 2 cases:

* when `withArgs($closure)` means a `I want this method to be called with 1 argument, which is closure`
* when `withArgs($closure)` means a `I want the closure to handle argument matching for me`
?

aik099:
Ah, the `withArgs` get's called inside `with` using `func_get_args()`. In that case, yes it's possible to make `withArgs` to accept a closure.

davedevelopment:
Yeah, `withArgs` as it stands is intended to take an array of all expected arguments, so we can just switch on type.

beni0888:
ok, I'll try it :)

peixinchen:
In my view, withArgs($argsOr$cloure) should be separated as withArgs(array $args) and withArgsUsing(Closure $closure), and it could solve the problom.

Revisor:
Yeah, just came here to say the same. The documentation for 0.9 regarding `withArgs()` doesn't match the code.

davedevelopment:
[...]

PR #545, opened 2016-02-02
Added new functionality to support `withArgs($closure)`to validate a list of arguments at once

beni0888:
This PR adds the new functionality requested in #459 . In this way, from now on it is possible to pass a closure to `withArgs()` method in order to validate a list of arguments at once. All the arguments passed in the call to the expected method will be passed to the closure, and if it evaluates to true, then the list of
arguments is assumed to have matched the expectation.

### DIFF
--- a/library/Mockery/Expectation.php
+++ b/library/Mockery/Expectation.php
@@ -285,15 +285,13 @@ class Expectation implements ExpectationInterface
             return true;
         }
         $expectedArgsCount = count($this->_expectedArgs);
-        if (count($args) !== $expectedArgsCount) {
-            if (
-                $expectedArgsCount == 1 &&
-                is_object($this->_expectedArgs[0]) &&
-                ($this->_expectedArgs[0] instanceof \Mockery\Matcher\MultiArgumentClosure) &&
-                $this->_matchArg($this->_expectedArgs[0], $args)
-            ) {
+        if ($expectedArgsCount === 1 && ($this->_expectedArgs[0] instanceof \Mockery\Matcher\MultiArgumentClosure)) {
+            if ($this->_matchArg($this->_expectedArgs[0], $args)) {
                 return true;
             }
+            return false;
+        }
+        if (count($args) !== $expectedArgsCount) {
             return false;
         }
         $argCount = count($args);
--- a/library/Mockery/Expectation.php
+++ b/library/Mockery/Expectation.php
@@ -284,7 +284,16 @@ class Expectation implements ExpectationInterface
         if (empty($this->_expectedArgs) && !$this->_noArgsExpectation) {
             return true;
         }
-        if (count($args) !== count($this->_expectedArgs)) {
+        $expectedArgsCount = count($this->_expectedArgs);
+        if (count($args) !== $expectedArgsCount) {
+            if (
+                $expectedArgsCount == 1 &&
+                is_object($this->_expectedArgs[0]) &&
+                ($this->_expectedArgs[0] instanceof \Mockery\Matcher\MultiArgumentClosure) &&
+                $this->_matchArg($this->_expectedArgs[0], $args)
+            ) {
+                return true;
+            }
             return false;
         }
         $argCount = count($args);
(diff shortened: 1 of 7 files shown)


library/Mockery/Expectation.php
```diff
@@ -331,7 +340,7 @@ protected function _matchArg($expected, &$actual)
         if ($expected instanceof \Mockery\Matcher\MatcherAbstract) {
             return $expected->match($actual);
         }
-        if (is_a($expected, '\Hamcrest\Matcher') || is_a($expected, '\Hamcrest_Matcher')) {
+        if ($expected instanceof \Hamcrest\Matcher || $expected instanceof \Hamcrest_Matcher) {
```
aik099: The `is_a` was used, because we're unsure if mentioned class is present. Probably now we can use `instanceof` because Hamcrest is a required dependency, but this isn't the topic of this PR.

beni0888: I'm sorry but I'm afraid that I don't see your point. I don't know what's the difference between using `is_a()` and using `instanceof` about this issue. If the target class wasn't present, `instanceof`would just return `false`, just like `is_a()` does, so I can't see the problem here. 

I agree with you about this isn't the topic of this PR, but I like to apply the boy scout rule when it is possible and I thought this was a good chance. Anyway, if there is any problem with that I can leave that commit out of this PR.

aik099: I like `boy scout rule`, but it actually hurts code reviews when used with combination of PR sending IMO, because it introduces changes not related to PR topic.

You can send another PR with just these changes (`is_a` change to `instanceof`).

davedevelopment: I think we get the idea, don't worry about issuing another PR, just consider it next time.

beni0888: @aik099 :+1:  you're right, it makes code reviews more difficult. I'll consider it next time :)

aik099: Thanks @beni0888 .

library/Mockery/Matcher/MultiArgumentClosure.php
```diff
@@ -0,0 +1,54 @@
...
+     * Actual passed by reference to preserve reference trail (where applicable)
+     * back to the original method parameter.
+     *
+     * @param mixed $actual
+     * @return bool
+     */
+    public function match(&$actual)
+    {
+        $closure = $this->_expected;
+        $requiredParamsNumber = (new \ReflectionFunction($closure))->getNumberOfRequiredParameters();
```
davedevelopment: I'm not sure if we need this level of strictness here and we can avoid having to worry about trailing args etc

davedevelopment: (Very open to discussion on it)

beni0888: Yeah, I think you're right. I fact, it doesn't make any sense doing that verification here. In addition, it introduces a bug as it returns false if the number of received params is not exactly the same as the number of required params of the closure, so if the closure receives some of the optional params it's going to return false even if the param is valid. I'm going to fix it immediately. I feel so ashamed right now...

davedevelopment: > I feel so ashamed right now...

Can't tell if serious or not, so just in case, please don't feel any shame!!! This is an awesome contribution and this is what PRs are for

aik099: Agree with @davedevelopment opinion.

davedevelopment:
@beni0888 this looks great, could you do one last thing and add a quick one line summary to the `CHANGELOG.md` file please, under the 1.0.0 heading, then I'll get this merged.

Thanks

library/Mockery/Matcher/MultiArgumentClosure.php
beni0888: Thanks guys! I was just being a little bit dramatic :-). But I really felt a little bit stupid for not detect that mistake before...

beni0888:
@davedevelopment Of course, I'll add it immediately. I forgot that little detail

beni0888:
@davedevelopment It's done :)

davedevelopment:
Thanks!

PR #546, opened 2016-02-06
Aliases expectation for boolean values

kwreczycki:
Hi,
I think that aliases for booleans are more readable in tests. 
How about that ?


davedevelopment:
Thanks, looks fine to me!

Issue #549, opened 2016-02-13
Times expectation with float value as a parameter should throw an exception


TiagoMaiaL:
Guys, I have a code similar to this:

`$floatValue = 1.3;`
`$fakeObject = \Mockery::mock('class_name');`
`$fakeObject->shouldReceive('method_name')->times($floatValue);`
`$fakeObject->method_name();`

This gives me an exception message similar to this one: "Method method_name() from ClassName should be called exactly 1.3 times but called 1 times.

I believe this code should throw an exception explaining that using float numbers with the times expectation method is not possible.

PR #550, opened 2016-02-14
Throwing an exception when times() receives a non integer value as it's parameter.

TiagoMaiaL:
Fixing [#549](https://github.com/padraic/mockery/issues/549) issue.

### DIFF
--- a/library/Mockery/Expectation.php
+++ b/library/Mockery/Expectation.php
@@ -559,6 +559,7 @@ class Expectation implements ExpectationInterface
      * Indicates the number of times this expectation should occur
      *
      * @param int $limit
+     * @throws InvalidArgumentException
      * @return self
      */
     public function times($limit = null)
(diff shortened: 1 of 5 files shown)


library/Mockery/Expectation.php
```diff
@@ -566,6 +566,9 @@ public function times($limit = null)
         if (is_null($limit)) {
             return $this;
         }
+        if (!is_int($limit)) {
+            throw new Exception('The passed Times limit should be an integer value');
```
beni0888: I think the exception thrown here is too generic. In my opinion it'd be much more expressive to throw an `InvalidArgumentException`.

TiagoMaiaL: Good point, i'll change this right now.

davedevelopment:
Thanks!

PR #553, opened 2016-02-23
respect COMPOSER_VENDOR_DIRECTORY envar

zodiac:
Composer allows the `COMPOSER_VENDOR_DIR` environmental variable (https://getcomposer.org/doc/03-cli.md#composer-vendor-dir) to be set; if it is, it installs dependencies in that directory instead of in `vendor`. However `Bootstrap.php` will still look for dependencies and `autoload.php` in the `vendor` directory, and will not find it there.

### DIFF
--- a/tests/Bootstrap.php
+++ b/tests/Bootstrap.php
@@ -31,11 +31,11 @@ error_reporting(E_ALL);
 $root    = realpath(dirname(dirname(__FILE__)));
 $library = "$root/library";
 $tests   = "$root/tests";
-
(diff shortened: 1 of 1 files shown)


davedevelopment:
Thanks!

beni0888:
@zodiac @davedevelopment I think that there is a bug in the already merged solution to this issue. The point is that `COMPOSER_VENDOR_DIR` can store both relative and absolute paths to the desired directory and in the implemented solution you're assuming that it's always a relative path.

PR #556, opened 2016-03-01
Refactor issue 459

beni0888:
This pull request introduce only refactoring changes, it doesn't modify or introduce any new functionality. I've made it because some time ago I was doing a code review with a friend over my previous pull request and I realized that there were some changes that there were some changes that could be done on the code to increase its readability, understandability and maintainability.

### DIFF
--- a/library/Mockery/Expectation.php
+++ b/library/Mockery/Expectation.php
@@ -20,6 +20,9 @@
 
 namespace Mockery;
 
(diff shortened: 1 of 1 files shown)


davedevelopment:
Looks complicated, but the tests pass, so :shipit:

davedevelopment:
Thanks!

Issue #557, opened 2016-03-02
Fatal error when trying to use Mockery::namedMock() in a class with an abstract method


TiagoMaiaL:
Guys, when trying to reproduce the issue [525](https://github.com/padraic/mockery/issues/525), I found another little bug. NamedMock isn't capable of generating a method with the same parameter declarations as the specified superclass. It gives a fatal error:

Illustration and reproducible steps (Those are based on the ones within issue 525):

I've created a small file to contain dummy namespaces and classes definitions:
```PHP
namespace Nature
{
    class Plant
    {
    }
}

namespace
{
    abstract class Gardener
    {
        abstract public function water(Nature\Plant $plant);
    }
}
```

In a new test method inside NamedMockTest:

```PHP
// require the dummy test classes.
// ...
$cactus = new \Nature\Plant();
$gardener = Mockery::namedMock(
    "NewNamespace\\ClassName",
    Gardener::class,
    array('water' => true)
);
$this->assertTrue($gardener->water($cactus));
```

The result of the previous test invocation is expressed in the following screenshot:
![screen shot 2016-03-02 at 12 00 52 pm](https://cloud.githubusercontent.com/assets/6910889/13464186/ac5eb81a-e06e-11e5-8a3f-e5941fa5af2a.png)

Sazpaimon:
Any chance of this getting merged? I just ran into this myself.

PR #558, opened 2016-03-02
Fixing issue #557

TiagoMaiaL:
Fixing issue [#557](https://github.com/padraic/mockery/issues/557).

Notes on the changes:

I've created a new file to include classes that are going to be used as dummy inside the tests. I've done this because I needed to use different namespaces, and to be able to do that, I needed to include a file called `Namespaced.php`.

In `MethodDefinitionPass.php` at line 86, I've created an array to hold the TypeHints to be used in the method definition. The problem with this is that the code is not maintainable neither reusable. I  researched in the project for similar arrays, and I found at least three uses of those in the following files:
* `MockConfigurationBuilder` at line 9 
* `KeywordPatch` at line 66. 

I suggest to put those arrays in a `functions.php` file or any other file that can make this code more reusable and maintainable.

### DIFF
--- a/tests/Mockery/NamedMockTest.php
+++ b/tests/Mockery/NamedMockTest.php
@@ -55,17 +55,6 @@ class NamedMockTest extends MockeryTestCase
     }
 
     /** @test */
-    public function itCreatesConcreteImplementationOfMethodsAndClasses()
-    {
-        $orchid = new \Nature\Plant();
-        $gardiner = Mockery::namedMock(
-            "GenericNamespace\\genericName",
-            Gardiner::class
-        );
-        $gardiner->water($orchid);
-    }
-
-    /** @test */
     public function itCreatesConcreteMethodImplementationWithReturnType()
     {
         $cactus = new \Nature\Plant();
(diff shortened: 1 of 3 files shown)


tests/Mockery/NamedMockTest.php
```diff
@@ -51,4 +53,16 @@ public function itShouldThrowIfAttemptingToRedefineNamedMock()
...
         $mock = Mockery::namedMock("Mockery\Dave7", "DateTime");
     }
+
+    /** @test */
+    public function itCreatesConcreteMethodImplementationWithReturnType()
+    {
+        $cactus = new \Nature\Plant();
+        $gardener = Mockery::namedMock(
+            "NewNamespace\\ClassName",
+            Gardener::class,
```
davedevelopment: I think the `class` constant is failing the tests, it looks like we're still supporting PHP 5.4. We should probably drop it, but for the sake of getting this merged you may as well make the change here.

TiagoMaiaL: Understood, i'm going to remove this right now.

Sazpaimon:
Any chance of this getting merged? I just ran into this myself.

davedevelopment:
Apologies! Adding new commits to a PR doesn't send any notifications, so I didn't know @TiagoMaiaL had addressed my comments...

Sazpaimon:
Thank you!

TiagoMaiaL:
No problems! Thanks for getting this merged.

PR #565, opened 2016-03-30
Fixing increased visibility issue(497).

TiagoMaiaL:
Fixing [issue 497](https://github.com/padraic/mockery/issues/497).

Notes about changes:

- I've changed the `mockery_getMethods` method in the `Mock` class. 
- The method now doesn't treat prototypes, it just lists the reflection methods of the current class.
- If there's a real reason for treating prototype methods, it's not being covered by the tests.
- I've changed the description of  an exception thrown when trying to mock a protected method, it's an attempt to be more accurate. The change can be seen in the `Mock.php` file, at line 207.
- I've removed unused commented code from `Mock` and `MockInterface` source.

### DIFF
--- a/library/Mockery/Mock.php
+++ b/library/Mockery/Mock.php
@@ -204,7 +204,7 @@ class Mock implements MockInterface
                         throw new \InvalidArgumentException("$method() cannot be mocked as it is a private method");
                     }
                     if (!$allowMockingProtectedMethods && $rm->isProtected()) {
-                        throw new \InvalidArgumentException("$method() cannot be mocked as it a protected method and mocking protected methods is not allowed for this mock");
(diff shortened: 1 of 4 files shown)


davedevelopment:
I imagine the `getPrototype` is something to do with abstract classes, will do some digging when I get chance to make sure we aren't regressing something there.

TiagoMaiaL:
Ok, if we're indeed regressing something, I can write a test that covers the code using this prototype in the expected way, so we could prevent this regression in the future. Also, if the regression is happening,  I could also solve the problem in another way. So, when it becomes possible for you, please, let me know about the results.

davedevelopment:
Thanks! I think the `_mockery_handleMethodCall` method already handles the abstract thing I mentioned.

TiagoMaiaL:
Nice! Thanks.

PR #566, opened 2016-04-01
first expectations are less specific than last

tflori:
a simple array_reverse can fix some tests at least in my case.

but it should be testet too. say me where I should write this test and I can write a test.


tflori:
I already realized it breaks other things and working on it

tflori:
why from master comes code with wrong code style? :-1:

robertbasic:
Hey @tflori thanks for the PR, and I apologize that no one from the team took a look at it yet. I'm trying to fix that now.

Could you explain what's the original issue that you're trying to solve with this PR?

Thanks!

tflori:
Hi, I don't remember which issue it was but at the end I solved my problems with `->byDefault()`.

Now that I know about by default I don't need it anymore but I still think that this order makes more sense.

davedevelopment:
I think we'd need to do something more complex if we wanted this as a feature and it would need to make it's way to 2.0, so will close for now.

PR #571, opened 2016-04-07
Adding missing docblock comments to the methods on Mockery.php

TiagoMaiaL:
Notes:
- Some methods still don't present some header description, when I get more time, this will be new a pull request. The same is true for other files that present this inconsistency.


davedevelopment:
Thanks!

PR #574, opened 2016-04-09
Mock's mockery_isAnonymous method now checks against implemented interfaces.

TiagoMaiaL:
Fixes [issue #529](https://github.com/padraic/mockery/issues/529).

Notes about changes:
- `mockery_isAnonymous` now checks if the current mock implements more interfaces rather than implementing only the `Mock` one.
- If, at some day, Mock start to implement more interfaces, this check will fail and will need to be corrected.

Suggestions:
- The Exception thrown when mocking non existent methods should be more specific, maybe with only a name specifying the situation.

### DIFF
--- a/library/Mockery/Mock.php
+++ b/library/Mockery/Mock.php
@@ -586,7 +586,8 @@ class Mock implements MockInterface
     public function mockery_isAnonymous()
     {
         $rfc = new \ReflectionClass($this);
-        return false === $rfc->getParentClass();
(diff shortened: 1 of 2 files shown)


davedevelopment:
@TiagoMaiaL could you rebase against master please?

TiagoMaiaL:
Of course, doing this right now.

TiagoMaiaL:
Done.

davedevelopment:
Thanks!

Issue #475, opened 2015-06-11
Introduce better error message for a method without expectations


bartmcleod:
If you mock *Zend\Stdlib\Hydrator\HydratorInterface* you get a mock object that mocks two interfaces: *Zend\Stdlib\Extractor\ExtractionInterface* and *Zend\Stdlib\Hydrator\HydrationInterface*, because the mocked interface inherits these two. This gives the mock object two methods: *hydrate* and *extract*. 

If you then call the *extract* method on the mock object, you get this error message: "Method Mockery_9_Zend_Stdlib_Hydrator_HydratorInterface::extract() does not exist on this mock object". This is vague, because the method exists, you just have to setup some expaction for it.

Calling *hydrate* works just fine.

davedevelopment:
I think something like "Received Mockery_9_Zend_Stdlib_Hydrator_HydratorInterface::extract(), but no expectations were specified" would be more appropriate.

We could take it one step further and spit out some example code, possibly with reflection magic.

bartmcleod:
I would stick to simple:  "Received a call to Mockery_9_Zend_Stdlib_Hydrator_HydratorInterface::extract(), but no expectations were specified."

aik099:
@davedevelopment , what exception will be triggered and with what message when method doesn't exist on mock object?

aik099:
I'm afraid that it's a single exception in both cases (existing method, but without expectations or non-existing method).

davedevelopment:
@aik099 definitely, we should have separate messages

bartmcleod:
The message when the method doesn't exist is fine, that can stay the same. A new message should be introduced, as proposed above, for the situation where the method exists but no expectations are defined.

PR #576, opened 2016-04-10
Adding specific message when calling method without any expectations on mock.

TiagoMaiaL:
Fixes [issue #475](https://github.com/padraic/mockery/issues/475).


davedevelopment:
Awesome, thanks

PR #578, opened 2016-04-12
Update example to call mockery close

kazu9su:
Now, this test does not fail even if I change test case like below.
So I think it is better to update the example.

```php
//does not fail even if call once
$mock->shouldReceive('disengageWarp')->never()->ordered();
```


davedevelopment:
Thanks!

PR #579, opened 2016-04-13
allow mocking exceptions under hhvm

zodiac:
This makes the test introduced in https://github.com/padraic/mockery/pull/456/files pass under hhvm and adds a test for mocking subclasses of Exception.


davedevelopment:
Thanks!

PR #580, opened 2016-04-15
Clear the _filesToCleanUp array after unlink.

friek:
When multiple unit tests are run without process isolation, \Mockery::close will attempt to unlink files from the _filesToCleanUp array every time.
This patch will clear the _filesToCleanUp array after the unlink loop, so no new attempts will be made to clean up files which don't exist anymore anyway.


davedevelopment:
Thanks!

Issue #564, opened 2016-03-28
php segfaults when mocking class with __call


zodiac:
php 5.5.33 segmentation faults when running the following code

```
<?php

require('/path/to/vendor/autoload.php');

class CommandRunner {
    public function __call($name, $args) {
        return 1;
    }
}

$cr = Mockery::mock('CommandRunner')->makePartial();
$cr->fopen("/tmp/nonexistent", 'r');
```

I know mockery doesn't support passing through to non-real methods (http://docs.mockery.io/en/latest/reference/mockery/gotchas.html) but a segfault seems excessively harsh

davedevelopment:
Confirmed, not been able to run with a debugger yet.

robertbasic:
The PR got merged, so this issue can be closed @davedevelopment!

robertbasic:
Closing as #584 was merged.

alexweissman:
It would appear that this has not been completely resolved - see #777.  I am using the latest release of Mockery (0.9.9).

alexweissman:
Nevermind, just realized that 1.0.0-alpha existed!  Seems to have fixed the issue.

PR #584, opened 2016-04-23
Fix segfaults with a mocking class with __call

nobuf:
I confirmed #564 with php 5.6.20 & xdebug. It says:

```
PHP Fatal error:  Maximum function nesting level of '256' reached, aborting! in /Users/nobu/projects/mockery/library/Mockery/Loader/EvalLoader.php(34) : eval()'d code on line 340
```

In the Mockery_0_test_Mockery_TestWithMethodOverloading:

```php
public function __call($method, $args) // line 340
{
    return $this->_mockery_handleMethodCall($method, $args);
}
```

This pull request attempts to avoid calling non-existing method infinitely and throws `BadMethodCallException` as the document says. Hope it helps.

### DIFF
--- a/library/Mockery/Mock.php
+++ b/library/Mockery/Mock.php
@@ -742,7 +742,8 @@ class Mock implements MockInterface
 
         if (!is_null($this->_mockery_partial) && method_exists($this->_mockery_partial, $method)) {
             return call_user_func_array(array($this->_mockery_partial, $method), $args);
-        } elseif ($this->_mockery_deferMissing && is_callable("parent::$method") && !$this->hasMethodOverloadingInParentClass()) {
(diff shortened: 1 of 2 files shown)


davedevelopment:
Great, thanks!

GrahamCampbell:
What about parent of parent?

GrahamCampbell:
This doesn't work. recursion is needed here to properly fix this.

GrahamCampbell:
Oh, wait, I think this is ok, since the parent will inherit this stuff. My bad! Ignore me!

davedevelopment:
Roger roger

GrahamCampbell:
:troll:

PR #587, opened 2016-04-26
Remove TestListener

davedevelopment:


GrahamCampbell:
Well, if the next release is 1.0, why not just delete this all together? :trollface:

davedevelopment:
@GrahamCampbell that's what I did ;)

pschultz:
This has introduced a BC break that is rather hard to resolve. Removing the listener will now skip lots of mock assertions. Because of sebastianbergmann/phpunit#2261 that happens silently in phpunit <5.5.6 (2016-10-03).

I cannot stress this enough. This PR silently broke all phpunit test suites that use the listener. At the very least this change has to be clearly documented.

The suggested update for existing test cases seems to be to either extend MockeryTestCase or use the MockeryPHPUnitIntegration.

Extending MockeryTestCase is not always an option, since the test case may already extend something other than PHPUnit_Framework_TestCase. Testing frameworks seem to have developed the nasty habbit of coercing users to extend their test cases, which obviously doesn't work more than once. We always liked Mockery for not doing that, for JustWorking(tm).

Using the trait all the time is tedious and error prone because tests will not fail if one forgets to include the trait, even though the tests are not working as expected.

Why has this listener been removed? Any chance to get this PR reverted?

pschultz:
I just realized that hasn't been released yet, so I suppose my comment about the PR breaking existing test suites is unwarranted. However, my concern about the BC break still stands and I find the listener much more elegant than the trait.

davedevelopment:
@pschultz It was removed as it wasn't working properly, due to changes in PHPUnit. See: padraic/mockery#376 and sebastianbergmann/phpunit#1440. I totally agree regarding the Listener being more elegant and miss it every day :cry: 

As for it not breaking test suites, it didn't occur to me that PHPUnit would silently ignore a missing listener and frankly I'd be placing the fault with them, rather than us. 

We could probably fixed it by recreating the Listener and just have it throw an exception to warn people that it doesn't work and will be permanently removed in the not too distant future. Sound reasonable?

pschultz:
To clarify: the fact that phpunit doesn't fail when a listener doesn't exist is a bug and has been fixed in 5.5.6 (about two weeks ago). The problem is that this bug is so old that it affects a whole bunch of test suites out there.

I am going to copy our conversation into a new issue where it should have more visibility.

PR #593, opened 2016-04-30
Adding TypeHint to getMagicMethods(), on MagicMethodTypeHintsPass.

TiagoMaiaL:
This is just a small refactor, instead of checking against classes, I've added the interface(`TargetClassInterface`) as the type hint. Changes on `getMagicMethods()`, within `MagicMethodTypeHintsPass`.


TiagoMaiaL:
Seems like php code sniffer fails to parse methods that have a reserved word as a name. code inside `SemiReservedWordsAsMethods`.

Issue #561, opened 2016-03-16
Tag a new release


ruudk:
I really want to use the latest stuff (PHP 7 support) from master. 

0.9.4 is the latest release and is from april 2, 2015.

davedevelopment:
We're a little in limbo at the minute, the master branch is what will be 1.0, which are I aren't going to be tagging just yet. I'd suggest just using the master branch as is for now. Tagging another 0.9.x release would be little different.

michaelmoussa:
@davedevelopment another 0.9.x release would be very helpful, actually. :)

We've been running with `"mockery/mockery": "dev-master#971dcdfdc2239b3d516672139f4368c19f6c2060 as 0.9.5"` for 3 or 4 months, since that's the commit that added support for scalar typehints and return types, and we're on PHP 7. Would be nice to be able to just specify `^0.9.5`.

GrahamCampbell:
> @davedevelopment another 0.9.x release would be very helpful, actually. :)

That won't be possible because of the breaking changes.

GrahamCampbell:
The next has to bump the minor/major vesion.

michaelmoussa:
`0.10.0` would be fine too.

GrahamCampbell:
But it was promised that there would never be a 0.10.0 release. The next would be 1.0.0.

michaelmoussa:
I was not aware of that, but consider that there hasn't been a new release in nearly a year. As a long-time user and advocate of Mockery, I think I would prefer a `0.10.0` release for PHP 7 compatibility over an indeterminate `1.0.0` release date, wouldn't you?

SemVer _does_ allow for BC breaks in major version zero. Besides, Mockery is a dev dependency, not a Production. The BC break (which, again, is allowed by SemVer), is unlikely to unexpectedly break any Production systems because it's not a library that one is expected to use in Production anyway.

davedevelopment:
If someone wants to create a branch off 0.9.4 and back port the relevant commits, I'm happy to tag a release, I don't currently have the time.

michaelmoussa:
@davedevelopment I'm happy to do this, but I can't create branches on this repo. Do you want to create that initial branch off of 0.9.4, and I'll send a PR to it with the relevant cherry-picked commits? Maybe call it `php7-compat` or something...?

mvdstam:
👍  Please tag `0.9.5` if possible.

nicwortel:
[...]

Issue #499, opened 2015-08-18
PHP 7: Mocking classes with typed __isset/__call/__callStatic results in an error


Majkl578:
``` php
class Test {
    function __isset(string $foo) {}
}

Mockery::mock(Test::class);
```

leads to:

```
Warning: Declaration of Mockery_0_Test::__isset($name) should be compatible with Test::__isset(string $foo)
```

Same applies to return typehints:

``` php
class Test {
    function __isset($foo) : bool {}
}

Mockery::mock(Test::class);
```

leads to:

```
Fatal error: Declaration of Mockery_0_Test::__isset($name) must be compatible with Test::__isset($foo): bool
```

michaelmoussa:
@Majkl578 this appears to have been fixed by 971dcdf. You can use this in your project by adding `"mockery/mockery": "dev-master#971dcdfdc2239b3d516672139f4368c19f6c2060 as 0.9.5"` in your hosts file and running `composer update mockery/mockery` until a new release is done.

@davedevelopment would you mind tagging a new release with this? It's been a while and some folks are already working on applications designed for PHP 7.

Majkl578:
I think I used dev-master at the time when I reported this issue. But I can check again.

Majkl578:
Still broken in ee88cd0.

5outh:
What is the status of this? Running into the issues mentioned above regarding type-hinted methods as well.

TiagoMaiaL:
Guess it wasn't solved neither is being resolved.

davedevelopment:
I'll see if I can set some time aside later today, can't make promises though. If anyone else fancies a crack, I think the way to do it would be to add a `Mockery\Generator\StringManipulation\Pass\Pass`, that adds the type hints in.

TiagoMaiaL:
I can actually take some time to solve this. I'll take some time to solve this.

davedevelopment:
@TiagoMaiaL :+1: thank you! Mockery's code generation is pretty hacky as it just manipulates strings, so regexes are your friend. We do string manipulation mostly because it is fast.

TiagoMaiaL:
Nice, I'll certainly take this into account.

robertbasic:
Closing as #585 was merged.

Please reopen if this is still an issue for you.

evsign:
Still an issue. Signature like this is not pass.
```
public function __call(string $method, array $arguments) : RequestBuilder;
```
Work only `public function __call($method, array $arguments);`

[...]

PR #596, opened 2016-05-09
Backport commits for PHP 7 support

michaelmoussa:
[This pull request](https://github.com/padraic/mockery/commit/3dd65d5181a3ffe4e5e406e7725928358e5967c7) adds support for PHP 7's scalar typehints and return type declarations. This PR adds that functionality to `0.9.x`. The changes seem to be backwards compatible as far as I can tell, so tagging either a new `0.9.5` or a `0.10.0` would be acceptable per SemVer.

This addresses: #561 (and, I think, #499)

### DIFF
--- a/library/Mockery/Expectation.php
+++ b/library/Mockery/Expectation.php
@@ -217,6 +217,32 @@ class Expectation implements ExpectationInterface
         } elseif (count($this->_returnQueue) > 0) {
             return current($this->_returnQueue);
         }
+
+        $rm = $this->_mock->mockery_getMethod($this->_name);
+        if ($rm && version_compare(PHP_VERSION, '7.0.0-dev') >= 0 && $rm->hasReturnType()) {
+            $type = (string) $rm->getReturnType();
+            switch ($type) {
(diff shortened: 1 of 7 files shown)


library/Mockery/Expectation.php
```diff
@@ -217,6 +217,32 @@ protected function _getReturnValue(array $args)
...
+        $rm = $this->_mock->mockery_getMethod($this->_name);
+        if ($rm && version_compare(PHP_VERSION, '7.0.0-dev') >= 0 && $rm->hasReturnType()) {
+            $type = (string) $rm->getReturnType();
+            switch ($type) {
+                case '':       return;
+                case 'string': return '';
+                case 'int':    return 0;
+                case 'float':  return 0.0;
+                case 'bool':   return false;
+                case 'array':  return [];
```
davedevelopment: I think if we change this to `return array();` this will be ready to ship?

prisis: yes, you cant use [] for array in 5.3

PR #598, opened 2016-05-17
Alternative shouldReceive syntax

davedevelopment:
Not sure what people think about this, it just gets away from using strings for method names...

``` php
$mock->shouldReceive()->foo("bar")->andReturn(123);
```

### DIFF
--- a/library/Mockery/ExpectationBuilder.php
+++ b/library/Mockery/ExpectationBuilder.php
@@ -1,0 +1,21 @@
+<?php
+
+namespace Mockery;
+
+class ExpectationBuilder 
+{
+    private $mock;
+
+    public function __construct(MockInterface $mock)
+    {
+        $this->mock = $mock;
+    }
+
+    /**
+     * @return Mockery\Expectation
+     */
+    public function __call($method, $args)
+    {
+        return $this->mock->shouldReceive($method)->withArgs($args);
+    }
+}
--- a/library/Mockery/ExpectationBuilder.php
+++ b/library/Mockery/ExpectationBuilder.php
@@ -5,10 +5,12 @@ namespace Mockery;
 class ExpectationBuilder 
 {
     private $mock;
+    private $method;
 
-    public function __construct(MockInterface $mock)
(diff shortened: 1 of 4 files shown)


library/Mockery/Mock.php
```diff
@@ -178,10 +179,10 @@ public function mockery_init(\Mockery\Container $container = null, $partialObjec
      * @param string $methodName,... one or many methods that are expected to be called in this mock
      * @return \Mockery\Expectation
      */
-    public function shouldReceive($methodName)
+    public function shouldReceive($methodName = null)
```
GrahamCampbell: php doc needs updating to string|null

library/Mockery/ExpectationBuilder.php
```diff
@@ -0,0 +1,21 @@
...
+{
+    private $mock;
+
+    public function __construct(MockInterface $mock)
+    {
+        $this->mock = $mock;
+    }
+
+    /**
+     * @return Mockery\Expectation
```
GrahamCampbell: missing param annotations

davedevelopment:
``` php
<?php

$mock = Mockery::mock();
                                                 
$mock->shouldReceive()->foo(123)->andReturn(456);
$mock->shouldNotReceive()->bar(567);

var_dump($mock->foo(123));

$mock->shouldHaveReceived()->foo(123);
$mock->shouldNotHaveReceived()->bar(456);

Mockery::close();
```

davedevelopment:
Riffing on the same thing, I could leave the old school methods alone and do 

``` php
<?php

$mock = Mockery::mock();

// $mock->shouldReceive("foo")->with(123)->once()->andReturn(456);
$mock->expect()->foo(123)->andReturn(456);

// $mock->shouldReceive("bar")->with(567)->andReturn(789);
$mock->allow()->bar(567)->andReturn(789);
```

Might collide with some userland code. Not sure what equivalent I'd want to use for the spies.

usox:
+1
A bit late but I would really prefer having real methods instead of dealing with method names a string-parameters.

PR #601, opened 2016-06-01
Exclude SemiReservedWordsAsMethods.php on StyleCI

GrahamCampbell:
Sorry this file can't be fixed by StyleCI right now. We do hope to fix that soon, but in the meanwhile, this should be suffice.


davedevelopment:
Thanks Graham

PR #602, opened 2016-06-02
Updated the Magic Pass to use the proper standard for toString

thanosp:
`__toString` was not detected but `__tostring` was instead.
The test class was written was using the latter as well.


davedevelopment:
Cheers

PR #610, opened 2016-07-06
No need to export some files and folders

lucasmichot:
No need to export some files and folders

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,16 @@
+* text=auto
+
+/docs export-ignore
+/examples export-ignore
+/tests export-ignore
+/travis export-ignore
+.coveralls.yml export-ignore
+.gitattributes export-ignore
+.gitignore export-ignore
+.php_cs export-ignore
+.scrutinizer.yml export-ignore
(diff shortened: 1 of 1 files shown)


davedevelopment:
Is this something that people generally want? When I download a release tarball, I like that these things are there...

lucasmichot:
I supposed that when people want to work on this package they usually either clone or either install it with `--prefer-dist` ?

davedevelopment:
It's not necessarily about working on it for me, just generally.

That is, if I release a new version of Mockery, I'm releasing the code *and* everything that goes with it. 

I appreciate the "carbon footprint" effect of this, but I think that can be dealt with using proxies, like we do for a lot of things on the web. 

These type of changes have been made and then reverted on projects like Symfony and composer no longer recommends this type of thing in the docs.

davedevelopment:
I'm going to close this, I prefer to have all of the info as part of the tarball.

lichunqiang:
> I prefer to have all of the info as part of the tarball.

I agree, but I think should 'usefully' things. `tests` `travis` is useful for the user?

PR #612, opened 2016-07-06
Use PRETTY_PRINT syntax and reorder keywords

lucasmichot:
Use PRETTY_PRINT syntax and reorder keywords - avoid commit noise when adding/updating a keyword


davedevelopment:
Thanks!

PR #614, opened 2016-07-13
Clean-up temporary files on teardown

ricardofandrade:
When you mock a class that does not exist, or can't be found by autoloading,
Mockery creates a temporary file, so the class can be required. The clean-up
was left to the OS.

This change registers the files to be cleaned-up on teardown by Mockery.


davedevelopment:
Thank you!

PR #618, opened 2016-08-26
Fix: __debugInfo for xdebug

fr05t1k:
For issue: https://github.com/padraic/mockery/issues/617

### DIFF
--- a/library/Mockery/Generator/MockConfigurationBuilder.php
+++ b/library/Mockery/Generator/MockConfigurationBuilder.php
@@ -15,6 +15,7 @@ class MockConfigurationBuilder
         '__toString',
         '__isset',
         '__destruct',
+        '__debugInfo',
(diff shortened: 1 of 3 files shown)


davedevelopment:
I think it would be better to add `__debugInfo` to the list of [blacklisted methods](https://github.com/padraic/mockery/blob/master/library/Mockery/Generator/MockConfigurationBuilder.php#L26), that way people can still get their debugging info out of the mock as necessary. What do you think?

fr05t1k:
@davedevelopment do you mean to add it to $blackListedMethods propperty as default? I think it will be great.

fr05t1k:
@davedevelopment what do you think?

davedevelopment:
Looks great, sorry for the delay, been on holiday.

Thanks!

PR #621, opened 2016-09-15
[Docs] Enable php highlighting to php code

santigarcor:
This lines enable the php highlighting to the `code-block`s when the code doesn't start with `<?php`


davedevelopment:
Thanks!

PR #627, opened 2016-10-01
Elaborating on instance mocks

sepehr:
First, when I was reading the docs, it was very obscure to me what an instance mock does. Added a paragraph to the docs for better understanding of this concept.


davedevelopment:
Thanks!

PR #638, opened 2016-10-27
Restore the PHPUnit listener

### LINKED ISSUE #636: Missing TestListener does not fail in most phpunit versions.
Discussion continued from #587

@pschultz
> This has introduced a BC break that is rather hard to resolve. Removing the listener will now skip lots of mock assertions. Because of sebastianbergmann/phpunit#2261 that happens silently in phpunit <5.5.6 (2016-10-03).

> I cannot stress this enough. This PR silently broke all phpunit test suites that use the listener. At the very least this change has to be clearly documented.

> The suggested update for existing test cases seems to be to either extend MockeryTestCase or use the MockeryPHPUnitIntegration.

> Extending MockeryTestCase is not always an option, since the test case may already extend something other than PHPUnit_Framework_TestCase. Testing frameworks seem to have developed the nasty habbit of coercing users to extend their test cases, which obviously doesn't work more than once. We always liked Mockery for not doing that, for JustWorking(tm).

> Using the trait all the time is tedious and error prone because tests will not fail if one forgets to include the trait, even though the tests are not working as expected.

> Why has this listener been removed? Any chance to get this PR reverted?

@pschultz
> I just realized that hasn't been released yet, so I suppose my comment about the PR breaking existing test suites is unwarranted. However, my concern about the BC break still stands and I find the listener much more elegant than the trait.

@davedevelopment 
> @pschultz It was removed as it wasn't working properly, due to changes in PHPUnit. See: padraic/mockery#376 and sebastianbergmann/phpunit#1440. I totally agree regarding the Listener being more elegant and miss it every day :cry:

> As for it not breaking test suites, it didn't occur to me that PHPUnit would silently ignore a missing listener and frankly I'd be placing the fault with them, rather than us.

> We could probably fixed it by recreating the Listener and just have it throw an exception to warn people that it doesn't work and will be permanently removed in the not too distant future. Sound reasonable?

pschultz:
The listener has been removed in ede143167ae34e because it is called too
late in the test lifecycle. This marks tests incorrectly as risky, if
there are no other assertions in the test.

Simply removing the listener silently breaks existing testsuites for
PHPUnit <5.5.6, which will not complain about missing listeners.

The listener added in this commit will no longer verify mock assertions.
Instead, the listener will mark tests as failed if they don't verify the
mock assertions themselves.

This will surface the change for existing users of the listener and help them
migrate to the trait or base class. Is also serves as an opt-in safefail for
new Mockery users.

This is how it looks in action:

```
$ bin/phpunit 
PHPUnit 5.3.5 by Sebastian Bergmann and contributors.

...........F

Time: 595 ms, Memory: 43.25MB

There was 1 failure:

1) Classmarkets\Tests\Search\AnalysisRunnerTest::testRunAnalysis
Mockery's expectations have not been verified. Make sure that \Mockery::close() is called at the end of the test. Consider using Mockery\Adapter\Phpunit\MockeryPHPUnitIntegration or extending Mockery\Adapter\Phpunit\MockeryTestCase.

/home/pschultz/src/mockery/library/Mockery/Adapter/Phpunit/TestListener.php:42

FAILURES!
Tests: 12, Assertions: 45, Failures: 1.
```

Fixes #636

### DIFF
--- a/library/Mockery/Adapter/Phpunit/TestListener.php
+++ b/library/Mockery/Adapter/Phpunit/TestListener.php
@@ -1,0 +1,50 @@
+<?php
+/**
+ * Mockery
+ *
+ * LICENSE
+ *
+ * This source file is subject to the new BSD license that is bundled
+ * with this package in the file LICENSE.txt.
+ * It is also available through the world-wide-web at this URL:
+ * http://github.com/padraic/mockery/blob/master/LICENSE
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to padraic@php.net so we can send you a copy immediately.
+ *
+ * @category   Mockery
+ * @package    Mockery
+ * @copyright  Copyright (c) 2010 Pádraic Brady (http://blog.astrumfutura.com)
+ * @license    http://github.com/padraic/mockery/blob/master/LICENSE New BSD License
+ */
+
+namespace Mockery\Adapter\Phpunit;
+
+class TestListener extends \PHPUnit_Framework_BaseTestListener
+{
+    /**
+     * endTest is called after each test and checks if \Mockery::close() has
+     * been called, and will let the test fail if it hasn't.
+     *
+     * @param  PHPUnit_Framework_Test $test
+     * @param  float                  $time
+     */
+    public function endTest(\PHPUnit_Framework_Test $test, $time)
+    {
+        try {
+            // The self() call is used as a sentinel. Anything that throws if
+            // the container is closed already will do.
+            \Mockery::self();
+        } catch (\LogicException $_) {
+            return;
+        }
+
+        $e = new \PHPUnit_Framework_ExpectationFailedException(sprintf(
+            "Mockery's expectations have not been verified. Make sure that \Mockery::close() is called at the end of the test. Consider using %s\MockeryPHPUnitIntegration or extending %s\MockeryTestCase.",
+            __NAMESPACE__,
+            __NAMESPACE__
+        ));
+        $result = $test->getTestResultObject();
+        $result->addFailure($test, $e, $time);
+    }
+}
--- a/library/Mockery/Adapter/Phpunit/TestListener.php
+++ b/library/Mockery/Adapter/Phpunit/TestListener.php
@@ -31,6 +31,20 @@ class TestListener extends \PHPUnit_Framework_BaseTestListener
      */
     public function endTest(\PHPUnit_Framework_Test $test, $time)
     {
+        if (!$test instanceof \PHPUnit_Framework_TestCase) {
+            // We need the getTestResultObject and getStatus methods which are
+            // not part of the interface.
+            return;
+        }
+
+        if ($test->getStatus() !== \PHPUnit_Runner_BaseTestRunner::STATUS_PASSED) {
+            // If the test didn't pass there is no guarantee that
+            // verifyMockObjects and assertPostConditions have been called.
+            // And even if it did, the point here is to prevent false
+            // negatives, not to make failing tests fail for more reasons.
+            return;
+        }
+
         try {
             // The self() call is used as a sentinel. Anything that throws if
             // the container is closed already will do.
--- a/tests/Mockery/Adapter/Phpunit/TestListenerTest.php
+++ b/tests/Mockery/Adapter/Phpunit/TestListenerTest.php
@@ -1,0 +1,74 @@
+<?php
+/**
+ * Mockery
+ *
+ * LICENSE
+ *
+ * This source file is subject to the new BSD license that is bundled
+ * with this package in the file LICENSE.txt.
+ * It is also available through the world-wide-web at this URL:
+ * http://github.com/padraic/mockery/blob/master/LICENSE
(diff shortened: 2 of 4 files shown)


tests/Mockery/Adapter/Phpunit/TestListenerTest.php
```diff
@@ -0,0 +1,74 @@
...
+ * to padraic@php.net so we can send you a copy immediately.
+ *
+ * @category   Mockery
+ * @package    Mockery
+ * @subpackage UnitTests
+ * @copyright  Copyright (c) 2010 Pádraic Brady (http://blog.astrumfutura.com)
+ * @license    http://github.com/padraic/mockery/blob/master/LICENSE New BSD License
+ */
+
+class Mockery_Adapter_Phpunit_TestListenerTest extends PHPUnit_Framework_TestCase
```
pschultz: Let me know if you want me to refactor to namespaces. The existing tests didn't have any, so I stuck to that style.

library/Mockery/Adapter/Phpunit/TestListener.php
```diff
@@ -0,0 +1,50 @@
...
+     * @param  float                  $time
+     */
+    public function endTest(\PHPUnit_Framework_Test $test, $time)
+    {
+        try {
+            // The self() call is used as a sentinel. Anything that throws if
+            // the container is closed already will do.
+            \Mockery::self();
+        } catch (\LogicException $_) {
+            return;
```
pschultz: If there are is a better way to check this I'm all ears.

library/Mockery/Adapter/Phpunit/TestListener.php
```diff
@@ -0,0 +1,50 @@
...
+        }
+
+        $e = new \PHPUnit_Framework_ExpectationFailedException(sprintf(
+            "Mockery's expectations have not been verified. Make sure that \Mockery::close() is called at the end of the test. Consider using %s\MockeryPHPUnitIntegration or extending %s\MockeryTestCase.",
+            __NAMESPACE__,
+            __NAMESPACE__
+        ));
+        $result = $test->getTestResultObject();
+        $result->addFailure($test, $e, $time);
```
pschultz: This should probably only happen if there are expectations to begin with, right?

library/Mockery/Adapter/Phpunit/TestListener.php
davedevelopment: I think it will suffice. I guess you could argue that you might be using Mockery for stubs or spies and post verification isn't required, therefore this would scream at you without justification, but I think we'll worry about that if someone mentions it.

library/Mockery/Adapter/Phpunit/TestListener.php
davedevelopment: Ah, I guess my reply to the other question is more relevant here. We can make a choice, throw anyway as a way of warning, or try and be clever and call Mock::close(), work on the result of that call. I don't really mind either way and will go with your judgement.

davedevelopment:
@pschultz replied to all comments, I'm happy to merge as is, please make changes regarding my comments if you would like to, otherwise let me know and I'll merge.

pschultz:
Let's merge it as is. I'm updating one of our test suites now and it looks like tests will not fail if Mockery isn't involved.

I'll dig into Mockery's internals later to see if the listener can be improved. I've only ever used mocks, never spies and have to become familiar with the differences.

pschultz:
Actually hold on. There may be a bug. Let me finish migration our test suite. Then I'll be much more confident that this works correctly.

pschultz:
This is definitely broken. There are lots of cases where assertPostConditions isn't called but the listener is, for instance skipped and incomplete tests.

davedevelopment:
TLDR for spies, you don't set up *expectations* on them, they record all interactions and you manually verify against them afterwards, e.g. `$spy->shouldHaveReceived("foo")->with("bar")->twice()`;

pschultz:
Good to go now. I decided to not check unsuccessul tests at all.

As far as I can see, when using only spies it is not strictly necessary to call close(). But it also doesn't hurt so I agree to wait for complaints before getting clever in the listener.

The fact that assertPostConditions isn't called for skipped, incomplete and failing tests is actually a problem. I will create a new issue for that.

tests/Mockery/Adapter/Phpunit/TestListenerTest.php
```diff
@@ -0,0 +1,82 @@
...
+
+        $this->test->setTestResultObject($this->testResult);
+        $this->testResult->addListener($this->listener);
+
+        $this->assertTrue($this->testResult->wasSuccessful(), 'sanity check: empty test results should be considered successful');
+    }
+
+    public function testSuccessOnClose()
+    {
+        $mock = $this->container->mock('foo');
```
davedevelopment: Using `'foo'` here is causing some flaky tests to when running the full suite, just calling `mock()` for these purposes should be fine. Same on lines 56 and 71.

pschultz:
Let me know if you want to rebase into a single commit.

davedevelopment:
Thanks, great work.

Issue #645, opened 2016-11-29
On PHP nightly (~17/11/2016 or later), isMultiArgumentClosureExpectation blows up


CyberiaResurrection:
It looks like a very recent nightly build of PHP has changed some of the count() function's semantics.  As outlined here, https://wiki.php.net/rfc/counting_non_countables, calling count(null) now throws some sort of warning (and causes PHPUnit tests to thus blow up).

As this is on course to become part of PHP 7.2, I feel this is a bug and should be fixed as soon as possible.

Steps to reproduce:
1 - Set up a ->withNoArgs() mock expectation.
2 - Run test under a very recent PHP nightly (later than approx 17/11/2016 - https://github.com/php/php-src/commit/b476d2196214e3c9acf22d4ac7b2ef1599f5077b)
3 - Kaboom

Example travis build where I tripped over this:
https://travis-ci.org/Algo-Web/POData-Laravel/jobs/179711657

PR #646, opened 2016-11-29
Fix withNoArgs() to work with PHP 7.2

CyberiaResurrection:
Simplest way I could think of of fixing #645 .

### DIFF
--- a/library/Mockery/Expectation.php
+++ b/library/Mockery/Expectation.php
@@ -421,7 +421,7 @@ class Expectation implements ExpectationInterface
     public function withNoArgs()
     {
         $this->_noArgsExpectation = true;
-        $this->_expectedArgs = null;
+        $this->_expectedArgs = [];
         return $this;
     }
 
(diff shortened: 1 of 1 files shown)


library/Mockery/Expectation.php
```diff
@@ -330,7 +330,8 @@ protected function _matchArg($expected, &$actual)
         }
         if (is_string($expected) && !is_array($actual) && !is_object($actual)) {
             # push/pop an error handler here to to make sure no error/exception thrown if $expected is not a regex
-            set_error_handler(function () {});
+            set_error_handler(function () {
+            });
```
GrahamCampbell: that makes no difference for php 7.2

GrahamCampbell: ah, I see :)

CyberiaResurrection: I fixed it because styleCI complained, no other reason

davedevelopment:
Thanks!

c-harris:
I have rerun the travis project that tests this feature, and I am happy to report that this has resolved it.
https://travis-ci.org/Algo-Web/POData-Laravel/jobs/179722258

PR #647, opened 2016-11-29
Returns null for auto return value of void methods

### LINKED ISSUE #640: Void class is generated in /tmp - PHP 7.1
Hi,

I'm running on master and PHP 7.1. During mocking void methods I get error:
**Fatal error: Cannot use 'void' as class name as it is reserved in /tmp/MockeryGlSZOl on line 1**.
The content of the file:

``` php
<?php class void {}
```

This issue is probably related to my previous closed one #635 

Thanks

davedevelopment:
Fixes #640


prisis:
Hey :), can you make a new release with this fix?

PR #656, opened 2016-12-19
Clear the _filesToCleanUp array after unlink.

tfrommen:
See 00f4baf148c9daeac172b89672a9d7baa7036e51.


davedevelopment:
@tfrommen see the travis failure, the 0.9 branch still supports PHP 5.3 so you'll need to use `= array();`

davedevelopment:
Thanks!

PR #585, opened 2016-04-25
Adding new Pass that solves conflicts in the declaration of magic methods.

TiagoMaiaL:
Fixes [issue 533](https://github.com/padraic/mockery/issues/533) and [issue 499](https://github.com/padraic/mockery/issues/499#issuecomment-208865086).

Notes about changes:

- I've mostly added new code.
     - Addition of new tests for reproducing the scenarios in the issues(`MockingParameterAndReturnTypesTest`). 
     - Addition of a new Pass, in charge of resolving the method definition conflicts occurred between Mock's magic methods and the class being mocked(`MagicMethodTypeHintsPass`).
     - Addition of a new Test for this class(`MagicMethodTypeHintsPassTest`).
     - Adding return statement on the mock's `__isset` method.

The classes are not perfect, so discussions are welcome.

PR #662, opened 2016-12-30
Apply MagicMethodTypeHintsPass to interfaces

### LINKED ISSUE #661: PHP 7: Mocking interface with __toString() with return typehint causes Fatal erro
Taking the failing coder reported in #533 and modifying the definition from `class Printable` to `interface Printable` results in the following error:

PHP Fatal error:  Declaration of Mockery_0_Printable::__toString() must be compatible with Printable::__toString(): string in /vagrant/vendor/mockery/mockery/library/Mockery/Loader/EvalLoader.php(34) : eval()'d code on line 25

This is due to the [MagicMethodTypeHintPass only looking at the `targetClass`](https://github.com/TiagoMaiaL/mockery/blob/bb89b9ceb124d4dadb0a161e1489710c269cd65e/library/Mockery/Generator/StringManipulation/Pass/MagicMethodTypeHintsPass.php#L60) which is `null` when mocking an interface.

nCrazed:
Resolves #661 

Just expanded on the work done in #585 by making sure that MagicMethodTypeHintsPass looks at interfaces as well as the class and modifying the tests to look at both the class and the interface cases.

### DIFF
--- a/library/Mockery/Generator/StringManipulation/Pass/MagicMethodTypeHintsPass.php
+++ b/library/Mockery/Generator/StringManipulation/Pass/MagicMethodTypeHintsPass.php
@@ -57,6 +57,9 @@ class MagicMethodTypeHintsPass implements Pass
     public function apply($code, MockConfiguration $config)
     {
         $magicMethods = $this->getMagicMethods($config->getTargetClass());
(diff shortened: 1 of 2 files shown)


davedevelopment:
Looks great, thank you.

PR #664, opened 2017-01-06
Allow mocking nullable return types for 7.1

shieldo:
Port/ cherry-pick of https://github.com/padraic/mockery/commit/e393e2410985fd0eac1a3d1267cfe371a49bef5e (allowing nullable return types for PHP 7.1) to the 0.9 branch. Also includes support for using the `self` keyword in a return type.

### DIFF
--- a/library/Mockery/Generator/Method.php
+++ b/library/Mockery/Generator/Method.php
@@ -26,7 +26,17 @@ class Method
     public function getReturnType()
     {
         if (version_compare(PHP_VERSION, '7.0.0-dev') >= 0 && $this->method->hasReturnType()) {
-            return (string) $this->method->getReturnType();
+            $returnType = (string) $this->method->getReturnType();
+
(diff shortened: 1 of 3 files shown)


GrahamCampbell:
I think this was intentionally not put on 0.9, because new features need to go to 1.0?

davedevelopment:
@GrahamCampbell I just haven't been bothering with 0.9, happy for others to make contributions if they want to stick with the tagged releases. 

This will need updating to exclude the `::class` syntax to pass on 5.4 etc

shieldo:
Yeah I'd suggest this isn't a "new feature" per se as it's just keeping equivalent feature parity (i.e. expecting mocking to work no matter the type declarations) with a new version of PHP without dropping anything older.

Will fix up the tests re the `::class` syntax.

davedevelopment:
Thanks!

PR #668, opened 2017-01-17
allows() and expects() syntax

davedevelopment:
Further to #598 

## `expects()`  for setting up mock expectations

``` php
<?php

// old
$mock->shouldReceive("foo")->with(123)->andReturn(true)->once();
// or
$mock->shouldReceive()->foo(123)->andReturn(true)->once();

// new
$mock->expects()->foo(123)->andReturns(true);

```

## `allows()` for stubbing methods

``` php
<?php

// old
$stub->shouldReceive("foo")->with(123)->andReturn(true);
// or
$stub->shouldReceive()->foo(123)->andReturn(true);

// new
$stub->allows()->foo(123)->andReturns(true);

```

Despite the new methods on the mocks, this is fully BC as they are skipped if the class to be mocked contains matching methods.

I personally prefer this syntax, but I haven't put it to much use. It's always bothered me that `shouldReceive()` starts with `should`, when it doesn't actually form a mock expectation without the additional call to `->once()` or one of the other count methods.

My objectives here are not to remove the existing methods, but rather try and encourage the new methods going forward.

Todo

- [x] Basic docs in README
- [x] No args vs not caring about args

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 * Removed object recorder feature
 * Bumped minimum PHP version to 5.6
 * `andThrow` will now throw anything `\Throwable`
+* Adds `allows` and `expects` syntax
  
 ## 0.9.4 (XXXX-XX-XX)
 
--- a/README.md
+++ b/README.md
@@ -102,9 +102,9 @@ $double = Mockery::mock(BookRepository::class);
 $double->expects()->add($book);
 ```
 
-During the test, Mockery will ensure that the `add` method is called no more
-than once. After you have finished exercising the system under test, you need to
-tell Mockery to check that the method was called at all using the
+During the test, Mockery accept calls to the `add` method as prescribed.
+After you have finished exercising the system under test, you need to
+tell Mockery to check that the method was called as expected, using the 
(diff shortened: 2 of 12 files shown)


davedevelopment:
One thing I need to clarify is the behaviour with regards to matching no arguments or specifying that we don't care about the arguments.

davedevelopment:
Regarding my todo item on no args vs any args, there's a slight difference in behaviour between the old and the new syntax. Previously:

``` php
$double->shouldReceive("foobar")->andReturn(123);

$double->foobar(); // int(123)
$double->foobar(123); // int(123)
$double->foobar(true, "some string", 5.4); // int(123)
```

With the new syntax:

``` php
$double->allows()->foo()->andReturns(123);

$double->foo(); // int(123);
$double->foo(123); // Uncaught Exception Mockery\Exception\NoMatchingExpectationException
```

That is, when we say `allows()->foo()`, we really mean that's all we are allowing. To make it act like `shouldReceive("foo")`, we need to explicitly tell it `withAnyArgs`. 

``` php
$double->allows()->bar()->withAnyArgs()->andReturns(123);

$double->bar(); // int(123);
$double->bar(123); // int(123)
```

I am fien with this, but I'm curious as to what other people think...

PR #671, opened 2017-01-18
Bumps phpunit version

davedevelopment:


davedevelopment:
Travis failures are probably due to php versions, #666 should fix that

PR #672, opened 2017-01-18
fix array_merge usage in MagicMethodTypeHintsPass::apply

villfa:
FYI, I found this bug thanks to [phpstan](https://github.com/phpstan/phpstan).


davedevelopment:
Great catch, thanks.

PR #675, opened 2017-01-21
Allowed mocking __clone method

### LINKED ISSUE #669: An interface specifying the magic __clone method cannot be mocked
```php
interface X {
    public function __clone();
}

Mockery::mock(X::class);
```

>Fatal error: Class Mockery_1_X contains 1 abstract method and must therefore be declared abstract or implement the remaining methods (X::__clone) in Mockery\Loader\EvalLoader.php(16) : eval()'d code on line 25

Bilge:
There does not seem to be any compelling reason why void magic methods such `__construct`, `__destruct` or `__clone` should not be mockable. However, this PR only seeks to add `__clone` support.

As a bonus, added test namespace (`MockeryTest`) to Composer autoloader to save manually including test fixtures (and other test files) in tests.

Resolves #669.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -27,6 +27,11 @@
     "autoload": {
         "psr-0": { "Mockery": "library/" }
     },
+    "autoload-dev": {
+        "psr-4": {
+            "MockeryTest\\": "tests/Mockery"
+        }
+    },
     "extra": {
         "branch-alias": {
             "dev-master": "0.9.x-dev"
--- a/library/Mockery/Generator/MockConfigurationBuilder.php
+++ b/library/Mockery/Generator/MockConfigurationBuilder.php
@@ -8,7 +8,6 @@ class MockConfigurationBuilder
     protected $blackListedMethods = array(
         '__call',
         '__callStatic',
-        '__clone',
         '__wakeup',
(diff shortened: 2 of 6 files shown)


davedevelopment:
As mentioned on IRC in regards to BC, anyone currently cloning a Mockery mock would have to fix up their tests:

``` php
<?php

class Foo {
    function __clone() {}
}

$double = Mockery::mock(Foo::class);

$second = clone ($double);
// PHP Fatal error:  Uncaught BadMethodCallException: Received Mockery_0_Foo::__clone(), but no expectations were specified in /Users/davem/src/mockery/library/Mockery/Loader/EvalLoader.php(34) : eval()'d code:774
```

You mentioned `__destruct`, doing the same for `__destruct` would have a similar behaviour. Any mock that mocks a type that defines `__destruct` would have to have a `$mock->shouldReceive('__destruct')` somewhere.

patzerr:
hi,
@davedevelopment 
the possibility to mock  `__construct` on **overloaded** classes without breaking other unittests, see the changes in `library/Mockery/Mock.php::_mockery_constructorCalled` in Pull Request #781

Issue #200, opened 2013-08-22
Test traits


ernestoalejo:
There's currently no way to mock traits with Mockery.

[For reference]
http://sebastian-bergmann.de/archives/906-Testing-Traits.html

BallisticPain:
Maybe even 2015?

davedevelopment:
Pull requests welcome :)

Bilge:
I think we could put this on the roadmap for 2025.

BallisticPain:
LOL @Bilge 
Yes @davedevelopment I assumed so. Unfortunately at this time I don't have personal time to work on this and in my professional time I was able to find a satisfactory way to test a trait. This is where I got the idea on how I was going to do it: http://carlosbuenosvinos.com/php-trait-testing/

Thanks!

Zizaco:
:+1:

Bilge:
Friendly reminder that it is now 2016.

agwidarsito:
+1

vokke-daniel:
+1,000,000

rewchelsea:
+1,000,001

davedevelopment:
He all, looking for feedback on #685

robertbasic:
As #685 was merged, I'm going to close this issue.

PR #685, opened 2017-01-29
Easy testing for traits

davedevelopment:
Thought I'd have a crack at this to see what it looked like, a few people have whined about it in #200. It's not really what I'd consider a responsibility of a test double framework, but we have the tools to do it...

``` php
<?php

Mockery::globalHelpers();

trait Foo {
    function foo() {
        return $this->doFoo();
    }

    abstract function doFoo();
}

$mock = mock(Foo::class);
$mock->allows()->doFoo()->andReturns('wooop');

var_dump($mock->foo());
```

- [x] Tests
- [x] Docs
- [x] Changelog

### DIFF
--- a/library/Mockery/Generator/MockConfiguration.php
+++ b/library/Mockery/Generator/MockConfiguration.php
@@ -42,6 +42,13 @@ class MockConfiguration
     protected $targetInterfaceNames = array();
 
     /**
+     * A number of traits we'd like to mock, keyed by name to attempt to
+     * keep unique
+     */
+    protected $targetTraits = array();
+    protected $targetTraitNames = array();
+
+    /**
      * An object we'd like our mock to proxy to
      */
     protected $targetObject;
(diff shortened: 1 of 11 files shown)


library/Mockery/Generator/MockConfiguration.php
```diff
@@ -467,6 +498,14 @@ protected function getAllMethods()
             $methods = array_merge($methods, $class->getMethods());
         }
 
+        foreach ($this->getTargetTraits() AS $trait) {
+            foreach ($trait->getMethods() as $method) {
+                if ($method->isAbstract()) {
```
robertbasic: It's been a while since I last looked at mockery's internals, but shouldn't you call `isTrait` here? If not, why do we need the `isTrait` method? I don't see it called anywhere.

davedevelopment: It doesn't need calling there and therefore you are right, we don't need it for this implementation. I put it in there thinking I would, will back it out.

davedevelopment:
I don't think this has any bearing on this functionality, but maybe someone else might think of something, there is an RFC currently under discussion to bring interfaces to traits - https://wiki.php.net/rfc/traits-with-interfaces.

PR #687, opened 2017-01-30
Forces the return value to null

### LINKED ISSUE #686: Mocked nullable return type returns a mock instead of null
```php
class Test {
	public function foo() : ?int {}
	public function bar() : ?DateTime {}
}

$mock = Mockery::mock(Test::class);
$mock->shouldReceive('foo')->andReturnNull();
$mock->shouldReceive('bar')->andReturnNull();

var_dump($mock->foo()); // mock of DateTime, should be NULL
var_dump($mock->bar()); // int(0), should be NULL
```

Found in: 3c90be8.

davedevelopment:
Fixes #686

### DIFF
--- a/library/Mockery/Expectation.php
+++ b/library/Mockery/Expectation.php
@@ -538,7 +538,7 @@ class Expectation implements ExpectationInterface
      */
     public function andReturnNull()
     {
-        return $this;
+        return $this->andReturn(null);
     }
 
     public function andReturnFalse()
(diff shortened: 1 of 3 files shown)


wired00:
Just wondering when there might be a release including the return this null value fix?

davedevelopment:
@wired00 I'm not sure yet. I'm pressing for a 1.0 release, for now I'd recommend just doing

``` php
$double->shouldReceive("foo")->andReturn(null);
```

PR #693, opened 2017-02-09
Add support for void methods (PHP 7.1+)

lcobucci:


lcobucci:
@davedevelopment maybe this coverage report should be generated using the latest PHP version instead...

lcobucci:
@davedevelopment thanks for the quick merge!

I'm having an issue with this that the test didn't cover: `PHP Fatal error:  Cannot use 'void' as class name as it is reserved in /tmp/Mockeryrku7ol on line 1` so mockery is considering `void` as an unexisting class and it's trying to create it. Do you have any idea where can I fix that?

davedevelopment:
Maybe this 4884c3a0?

lcobucci:
The structure is quite different on `0.9` but I think I've found, thanks.
Sending a new PR now 😉

Issue #376, opened 2014-09-18
phpunit integration


spdionis:
If a test function does not make any assert then phpunit complains about "risky test" and Phpstorm says "This test did not perform any assertions".

I have added Mockery's listener to phpunit.xml. I've also tried to do \Mockery::close() in the tearDown() method.

Mockery's expectations still work as intended. They break when they should. Currently i fix the issue adding $this->assertTrue(true); in every test.

Phpunit version is 4.2.6 and Mockery version is 0.9.2

gknapp:
I'm seeing this issue in PHPUnit 4.6.0 with Mockery 0.9.4 - if I just use a mock Doctrine DBAL connection and have no assertions in the test, it's marked as risky with `beStrictAboutTestsThatDoNotTestAnything="true"` in my phpunit.xml

aik099:
Yes, mockery's mock expectation don't add-up to assertion count made in the test. That's why test is marked as risky. It's impossible to do this via PHPUnit hook. What you want is to call `\Mockery::close();` in your `tearDown` method or just use the proposed trait in your unit test.

halaei:
Using the proposed trait, this test is not marked as risky, although it really tests nothing:
```php
function test_nothing()
{
    $c = Mockery::mock(C::class);
    $c->shouldReceive('f');
}
```
Is there a fix for it as well?

XedinUnknown:
Still experiencing this with PHPUnit 6, and Mockery 1.2. Adding `Mockery::close()` or the `addToAssertionCount()` thing didn't work: still marks tests as risky.

XedinUnknown:
Fixed. Apparently, gotta just use the `Mockery\Adapter\Phpunit\MockeryPHPUnitIntegration` trait in the test case.

reza305z:
experiencing it too with shouldReceive() method, but not with shouldNotReceive(). Using PHPUnit 9, and Mockery 1.4. Adding Mockery\Adapter\Phpunit\MockeryPHPUnitIntegration trait in test case didn't work. I'm using Pest 1.16 by the way.
Only thing that can help is adding this method:

``
test('some_test', function() {

    $this->expectNotToPerformAssertions();
    $this->someMockedInterface->shouldReceive('someMethod');
    // ...

});
``

mfn:
This is because of the latest version: your `shouldReceive` has no _actual_ expectation set yet. It needs `->once()` or `->times()`. Or, instead of `shouldReceive`, you can use `expects`.

[...]

PR #695, opened 2017-02-10
Fix TestListener for php 6 version.

bruli:


davedevelopment:
Will take a look at this next week, my preference would be to continue supporting PHPUnit 5 alongside 6 for the time being.

pavarnos:
Will there be a mockery release soon? I'd like to migrate to phpunit 6 but can not until mockery is compatible...

davedevelopment:
My understanding is the master branch should now support PHPUnit 5 + 6, I don't know when there will be a release.

pavarnos:
The mockery TestListener adapter uses the old _ class name rather than the new namespaces?

davedevelopment:
My mistake! I didn't realise this was for the test listener - reopened. Regardless, the test listener shouldn't be a blocker for using PHPUnit 6, the test listener itself no longer functions properly and you should use the `MockeryTestCase` base class or the trait to manage the tear down.

pavarnos:
Can there be some kind of cross check added to protect you from forgetting to inherit from MockeryTestCase or use the trait (eg when migrating a large collection of tests from the Listener as you suggest above)

So maybe a project test bootstrap file could call `Mockery::usingPhpUnit()` and throw an exception if `Mockery::close()` is not called by the time `endTest()` is called in the test listener?

spdionis:
Why doesn't the test listener function properly? I liked this option best for the integration. Is it not possible to update it?

davedevelopment:
@spdionis heya, it was discussed here #376 (which you opened) and sebastianbergmann/phpunit#1440

spdionis:
@davedevelopment hey, right, I forgot about that discussion, it was quite a while ago :). Apparently we'll have to use trait or base class after all.

theofidry:
What is the status for this PR?

davedevelopment:
@theofidry a mess?!

The listener can't support both PHPUnit 6 and 5.7, unless we have separate class definitions. PHPUnit 5.7 has a forward compatibility layer, but it only goes so far. We can't specify one class that is compatible with PHPUnit 5.7's `PHPUnit_Framework_TestListener` and PHPUnit 6's `PHPUnit\Framework\TestListener`.

Giving the listener isn't essential to development (it's only a fail safe now, you need to use the trait or base class), perhaps we should just drop support for the listener in PHPUnit 5.7? What do people think?

theofidry:
Hm dropping the listener is a bit of an issue, we are using it for Humbug and Infection is missing some tests due to the lack of support for PHPUnit 6 for it.

I would rather suggest to drop PHPUnit 5.* support completely and move forward. It's not like Mockery is gonna disappear for people stuck on PHPUnit 5.*.

davedevelopment:
> I would rather suggest to drop PHPUnit 5.* support completely and move forward. It's not like Mockery is gonna disappear for people stuck on PHPUnit 5.*.

That's true, but we're so close to a 1.0 that does support 5.7, I'd rather not throw it all away. I think I'll make the listener PHPUnit 6 only, but keep support for everything else.

theofidry:
That would be an acceptable compromise

davedevelopment:
Closing in favour of #786

PR #699, opened 2017-02-16
Call parent __callStatic method

imaben:
eg.

```
class MyBaseClass {
    public static function __callStatic($name, $args)
    {
        if ($name == 'foo1') {
            // todo
        }
    }
}
$mock = Mockery::mock('MyNameClass', MyBaseClass::class);
MyNameClass::foo1();
```

So, it's will throw \BadMethodCallException

### DIFF
--- a/library/Mockery/Mock.php
+++ b/library/Mockery/Mock.php
@@ -250,8 +250,8 @@ class Mock implements MockInterface
         return new ExpectsHigherOrderMessage($this);
     }
     // end method expects
-     
-     
+
+
     /**
      * Shortcut method for setting an expectation that a method should not be called.
      *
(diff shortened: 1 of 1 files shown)


davedevelopment:
I can't think if there will be any wtfs with this change. The problem is, we allow people to set up expectations in a weird way, that is using the returned mock object to set up stubs and expectations for static calls, e.g.

``` php
 $mock = Mockery::namedMock('MyNameClass', MyBaseClass::class);
 $mock->allows()->foo1()->andReturns('dave');
```

This PR is effectively going to remove that checking of calls have been set up properly. Perhaps the `makePartial` functionality should also apply to static calls?

davedevelopment:
Just to be clear, I wasn't saying "no" to this PR, I was just opening up some concerns for discussion...

PR #703, opened 2017-02-21
replacing func_get_args with variadic parameters for better IDE-support

MarcusSchwarz:
Methods like "shouldReceive" may receive an arbitrary number of arguments. Yet according to the methods' signature no arguments are allowed. So many IDEs like PHPStorm assume a bug when the method is used as intended. The solution to this problem are variadic parameters, which are available since php5.6. I assume that at least 5.6 is required for the current master of Mockery. If 5.6 is the minimum requirement of the current stable version as well I will provide a patch for that version, too.

### DIFF
--- a/library/Mockery.php
+++ b/library/Mockery.php
@@ -72,12 +72,12 @@ class Mockery
     /**
      * Static shortcut to \Mockery\Container::mock().
      *
+     * @param array $args
+     *
      * @return \Mockery\MockInterface
      */
-    public static function mock()
+    public static function mock(...$args)
     {
-        $args = func_get_args();
-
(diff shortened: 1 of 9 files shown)


MartialGeek:
The variadic parameters are available since PHP 5.6, so the composer.json of Mockery must be updated to be compatible with this feature.
IMHO the safer solution to keep the backward compatibility is to add a type $args = null to those methods.

davedevelopment:
Thanks!

@MartialGeek We already have 5.6 as a minimum in the master branch

MartialGeek:
OK so I suppose this improvement is for your next major version. Thank you for your work ;)

PR #704, opened 2017-02-24
Fixing performance issue with uniqid

asuth:
I was doing some profiling and found that 15% of all of our test time was taken in `uniqid`. This removes the usage of uniqid because it should not be truly needed. This caused this part of the code to go from taking 97 seconds to taking 0.2 seconds.

Note: Our tests are making 50k calls to `hasMethodOverloadingInParentClass`, via `_mockery_handleMethodCall` 

This is also when run on HHVM 3.14.5. Mainline PHP may or may not have a faster implementation.

## Before
screen shot 2017-02-23 at 5 25 15 pm

## After
screen shot 2017-02-23 at 5 25 25 pm

### DIFF
--- a/library/Mockery/Mock.php
+++ b/library/Mockery/Mock.php
@@ -834,7 +834,7 @@ class Mock implements MockInterface
     private function hasMethodOverloadingInParentClass()
     {
         // if there's __call any name would be callable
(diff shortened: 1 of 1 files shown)


davedevelopment:
:100: for the choice in method name :)

PR #705, opened 2017-02-24
Support nullable parameters

lcobucci:
This PR adds support for nullable types on parameters.

### DIFF
--- a/library/Mockery/Generator/Parameter.php
+++ b/library/Mockery/Generator/Parameter.php
@@ -87,9 +87,6 @@ class Parameter
      */
     public function isVariadic()
     {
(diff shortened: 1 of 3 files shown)


lcobucci:
@davedevelopment if everything is fine can you please release `v0.9.9`?

davedevelopment:
Done. I'm pretty much over supporting 0.9 with these tiny patches and releases though, so I'd suggest moving on to 1.0.0-alpha1 sooner rather than later.

lcobucci:
@davedevelopment thanks! I told that to the team and they will start to move to `v1.0.0-alpha1` ASAP.

PR #708, opened 2017-03-02
don't declare two classes in one file

tflori:


davedevelopment:
I didn't know about the [class in 5.4 that you mentioned](https://github.com/padraic/mockery/commit/d19b9bb3ea70d80f0800fe1c68098cf3a11dc601#commitcomment-21122239), so if we can agree to drop support for PHPUnit < 5.4, we can skip the separate class definitions.

I definitely do not want to remove the trait completely, that's a backwards step for me.

tflori:
https://github.com/sebastianbergmann/phpunit/blob/5.4/src/ForwardCompatibility/TestCase.php

For me: yes but I'm not maintainer of mockery. mockery requires php >= 5.6 and in this versions you can use the still supported phpunit 5.7. I don't know a reason for using php 4.8 other than dealing with php 5.3.

btw.: phpunit stops supporting php5.6 in february

PR #710, opened 2017-03-08
PhpStorm and magic __call methods

unglud:
Problem: PhpStorm not always can follow to __call method and inspector may yell that it can not find method.
before

Solution: Add PHPDoc with methods descriptions
after

### DIFF
--- a/library/Mockery/ExpectationInterface.php
+++ b/library/Mockery/ExpectationInterface.php
@@ -21,6 +21,16 @@
 
 namespace Mockery;
 
+/**
+ * @method Expectation once()
+ * @method Expectation zeroOrMoreTimes()
+ * @method Expectation twice()
+ * @method Expectation times(int $limit)
+ * @method Expectation never()
+ * @method Expectation atLeast()
+ * @method Expectation atMost()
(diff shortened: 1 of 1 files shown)


davedevelopment:
Will merge this as a quick fix, but I think the better way would be to fix the interface and anything that implements it. This would break BC, but it's a shallow break, ExpectationInterface can be marked as `@internal` going forward. 

I've actually written about making the experience better here, but my implementation at the time branched off in to a completely new project

Blog Post: http://davedevelopment.co.uk/2014/11/03/designing-a-dsl-with-interfaces.html
Toy Project: https://github.com/davedevelopment/sham

unglud:
I agree it's more a hack, then fix. I'll look at your links, and may be came to some more right way.
Thank you.

davedevelopment:
@unglud no problem, I wasn't suggesting you look at it, more a note for myself or anyone else who happens to be passing by. Obviously you are more than welcome to take a look, just want to be clear I'm not trying to push you in to anything :)

unglud:
You will not stop me make this better! :)

Issue #711, opened 2017-03-08
Allow closure spies


tflori:
Currently I write something like this when I want to test that a closure gets executed:

```php
class SomethingTest extends TestCase 
{
  public function testSomething()
  {
    $spy = [
      'callCount' => 0,
      'args' => []
    ];
    
    aFunction(function () use (&$spy) {
      $spy['callCount']++;
      $spy['args'][] = func_get_args();
    });

    self::assertSame(1, $spy['callCount']);
    self::assertSame(['what', 'ever'], $spy['args'][0]);
  }
}
```

dependent from what I want to test of course. I'm wondering that there is no spy. In jasmine for example your write:

```js
describe('something', function() {
  it('calls the callback', function() {
    var spy = jasmine.createSpy('callback');
    
    aFunction(spy);

    expect(spy).toHaveBeenCalledWith('what', 'ever');
  });
});
```

Sounds some bit more easy, readable and understandable.

Ok, thats not your problem - phpunit is differnt. But you could provide a spy. I just saw that there is a method `Mockery::spy()` but when I see it correct is again to create a Spy for an Interface. I want to spy a closure. Ok spy is taken and I guess you don't want to change the current behavior for spy.

My suggestion:

```php
class SomethingTest extends TestCase 
{
  public function testSomething()
  {
    $spy = \Mockery::createSpy('callback');
    
    aFunction($spy);

    $spy->shouldHaveBeenCalled()->with('what', 'ever')->once();
  }
}
```

davedevelopment:
I'd be open to providing something like this. We could code up something specific, or we could utilise the existing code and effectively do something like:

```
class CallableSpy {
    public function __invoke() {}
    public function shouldHaveBeenCalled() {
        return $this->shouldHaveReceived("__invoke");
    }
}

$mock = Mockery::spy(CallableSpy::class)->makePartial();
$mock(123);
$mock->shouldHaveBeenCalled()->with(123);
```

My preferred syntax would be something like 

```
Mockery::spy(function() {});
```

I think it's fairly clear that it's a spy for a callable, and we could also further it to make it partial spy etc.

```
Mockery::spy(function($number) { return $number+1; });
```

tflori:
yes, thats good for me too.

davedevelopment:
@tflori see #712, let me know your thoughts on there.

robertbasic:
Closing as #712 got merged.

PR #712, opened 2017-03-09
RFC: Callable spies

davedevelopment:
Here's a take on the request in #711

``` php
 $spy = spy(function($n) { return $n + 1;});

 array_map($spy, [1, 2]); // [2, 3]

 $spy->shouldHaveBeenCalled();
 $spy->shouldHaveBeenCalled()->twice();
 $spy->shouldHaveBeenCalled()->with(1)->once();
 $spy->shouldHaveBeenCalled()->with(2)->once();

 $spy->shouldHaveBeenCalled()->with(3); // throws...
```

# Todos/considerations

- [ ] Better error messages? Ideally we could identify that these are a special case and avoid complicating things showing the internals of `ClosureWrapper::__invoke` etc
- [ ] Make it available for mocking as well? It would mean coming up with some syntax or probably overloading `expects` and `allows` somehow. Might get too complicated.
- [ ] `shouldNotHaveBeenCalledWith(123, 456, 789)` ? I think I prefer this for specifying arguments with a `not`

### DIFF
--- a/library/Mockery.php
+++ b/library/Mockery.php
@@ -93,8 +93,6 @@ class Mockery
      */
     public static function spy(...$args)
     {
-        $args = func_get_args();
-
         if (count($args) && $args[0] instanceof \Closure) {
             $args[0] = new ClosureWrapper($args[0]);
         }
--- a/library/Mockery.php
+++ b/library/Mockery.php
@@ -26,6 +26,7 @@ use Mockery\Generator\StringManipulationGenerator;
 use Mockery\Loader\EvalLoader;
 use Mockery\Loader\Loader;
 use Mockery\Matcher\MatcherAbstract;
+use Mockery\ClosureWrapper;
 
 class Mockery
 {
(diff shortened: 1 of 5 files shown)


davedevelopment:
Also probably worth considering if it's even worth putting this in to mockery. It's simple enough to maintain I think, but the functionality isn't too difficult to roll on your own as you don't really need the code generation etc.

library/Mockery.php
```diff
@@ -92,6 +93,12 @@ public static function mock(...$args)
      */
     public static function spy(...$args)
     {
+        $args = func_get_args();
```
MarcusSchwarz: This line should not be necessary. ...$args is an array of all function arguments already.

davedevelopment: :+1:

robertbasic:
@davedevelopment what's the status with this? Does it require any more work?

davedevelopment:
@robertbasic no further progress. Error messages are not very friendly at the moment. Happy to merge as an experimental feature.

tflori:
sorry, I somehow missed this MR.

In the mean time I found a problem we would have with this implementation in laravel: they have a lot of functions where they only allow `\Closure`. But I guess in this case the user has to write his own closure around the wrapper.

So for me this is good

davedevelopment:
Merged, will leave undocumented as it's still considered experimental

PR #727, opened 2017-04-26
Spies return `null` by default from ignored (non-mocked) methods with nullable return type

asprega:
- Fix + related test
- Documented BC BREAK on changelog
- Added a bit of documentation on how default return types work for mocks related to methods with return types

### DIFF
--- a/tests/Mockery/MockingNullableMethodsTest.php
+++ b/tests/Mockery/MockingNullableMethodsTest.php
@@ -211,4 +211,12 @@ class MockingNullableMethodsTest extends MockeryTestCase
 
         $this->assertEquals(null, $double->nullableInt());
     }
+
+    /** @test */
+    public function it_returns_null_on_calls_to_ignored_methods_of_spies_if_return_type_is_nullable()
+    {
+        $double = \Mockery::spy(MethodWithNullableReturnType::class);
+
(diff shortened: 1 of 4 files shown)


tests/Mockery/MockingNullableMethodsTest.php
```diff
@@ -211,4 +211,12 @@ public function it_allows_returning_null_for_nullable_int_return_types()
...
         $this->assertEquals(null, $double->nullableInt());
     }
+
+    /** @test */
+    public function it_returns_null_on_calls_to_ignored_methods_of_spies_if_return_type_is_nullable()
+    {
+        $double = \Mockery::spy(MethodWithNullableReturnType::class);
+
+        $this->assertEquals(null, $double->nullableClass());
+    }
```
asprega: I did not add multiple tests as it looked overkill to me, but if you prefer I can add one for every specific return type contained in `MethodWithNullableReturnType`

davedevelopment:
Great, thank you.

asprega:
no prob! thanks for merging. Is there any (approx) ETA for a new tagged version, I assume 1.0.0?

davedevelopment:
@asprega Not at this time, I already have an alpha tag for 1.0.0, but I want to make sure we're super happy before I stamp that full 1.0.0.

Issue #728, opened 2017-04-26
PHPUnit 6 support


padraic:
Hey @davedevelopment, are there any blockers to this other than renaming test classes/ exception methods where appropriate in the tests? I see a few conflicts over on Humbug repos, I'd like to resolve. So happy to do a PR over the weekend.

davedevelopment:
No blockers I am aware of, fire away. I think we fixed one thing in #691 (and then reverted slightly in #708).

davedevelopment:
#708 says we're dropping support for PHPUnit <5.4, I'd like to try and keep compatibility with everything after that if we can.

padraic:
I'm assuming 5.4 is 5.6?

padraic:
Next question - how tied are we to supporting the TestListener post PHPUnit 5? The problem that I see is that it's method signature for endTest() is effectively dead after 5.7 so we can't cross-support it without building a new one. It would be easier to skip the tests for it in 6.0+.

davedevelopment:
I meant PHPUnit 5.4, I was told that they added `\PHPUnit\Framework\TestCase` in that release, so we've extended that one in the `MockeryTestCase` class.

I consider the listener as deprecated really, it's been turned in to some sort of fail safe for now. There is a PR to convert it to PHPUnit 6 (#695), but I was hoping to support both going forward, looks like that wont happen. Happy to see the tests skipped for the listener for now.

padraic:
I can try lowering it in Travis (well, try it here first!)- currently that's set to 5.7.

davedevelopment:
It depends which way you are looking at it though and it's hard to test! To me `require-dev` tells us the PHPUnit version that is required to run our test suite, not the PHPUnit version that Mockery attempts to be compatible with, if that makes sense?

padraic:
Ahh, understood. That is tricky to check. Running our own tests under 5.4 would at least pick out any class name misses (e.g. if the forward compatibility shim is less developed back then). At least, that's been the main problem moving require-dev to 6.1, and usually in the places where we do directly use phpunit classes.

ascii-soup:
Any movement on this?

davedevelopment:
I'm a bit lost on where we are with it to be honest, will see if I can review this week.

jasonetwork:
Any updates on getting Mockery compatible with the latest version of PHPUnit?

dbudwin:
[...]

PR #729, opened 2017-04-27
[WIP] PHPUnit 6 Support

padraic:
Initial run through of changes for #728. The main areas needing attention:
* PHPUnitConstraint - Incompatible between PHPUnit 5 and 6 as the referred to typehints and such are not part of the PHPUnit 5 forward compatibility shim.
* Deprecation error under PHP 7.0 (on one test where constructor shares class name).
* Numerous reports related "risky tests" under PHPUnit 6.0 though presumably that's all down to test framework in place which needs to be updated.
* The Listener is also incompatible with PHPUnit 6 (typehints again).

### DIFF
--- a/tests/Mockery/Matcher/PHPUnitConstraintTest.php
+++ b/tests/Mockery/Matcher/PHPUnitConstraintTest.php
@@ -75,11 +75,9 @@ class PHPUnitConstraintTest extends TestCase
         $this->assertTrue($this->rethrowingMatcher->match($value3));
     }
 
-    /**
-     * @expectedException \PHPUnit\Framework\AssertionFailedError
-     */
     public function testMatchesWhereNotMatchAndRethrowing()
     {
+        $this->expectException($this->assertionFailedError);
         $value = 'value';
         $this->constraint
             ->shouldReceive('evaluate')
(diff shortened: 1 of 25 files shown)


padraic:
@davedevelopment This is probably ready for review. A few notes:

1. Updated PHPUnitConstraint to switch between classes as needed.
2. TestListener tests are skipped under PHPUnit 6 - I also excluded them from code coverage since that would error with PHPUnit 6 runs also.
3. (Optional) I updated the Travis config to add a new build which run tests under PHP 5.6 using PHPUnit 5.4. Can't think of anything off hand other than that to check there are no obvious class name conflicts.

padraic:
Note: I haven't done anything about the risky tests which don't fail the Travis build under PHPUnit 6. Maybe a separate PR.

tests/Mockery/Matcher/PHPUnitConstraintTest.php
```diff
@@ -33,7 +34,18 @@ class PHPUnitConstraintTest extends \PHPUnit_Framework_TestCase
 
     public function setUp()
     {
-        $this->constraint = \Mockery::mock('PHPUnit_Framework_Constraint');
+        /*
+         * Revise by PHPUnit version
+         */
+        if (class_exists('\PHPUnit\Framework\AssertionFailedError')) {
```
GrahamCampbell: no need for a slash, php just deletes it
https://github.com/php/php-src/blob/d2285e1613f239043d439a6b439ee510452f5e07/Zend/zend_builtin_functions.c#L1376
https://github.com/php/php-src/blob/5340587c253fad8e6bbfbbf62372ba017970ea1a/Zend/zend_execute_API.c#L961

davedevelopment:
Looks all good to me @padraic

Issue #442, opened 2015-02-19
State of the documentation


robertbasic:
Given that we have now a couple of tickets labelled `Towards 1.0.0` I thought to ask about the state of the documentation.

What new features were added that need docs? I see the spy stuff is new (to me) https://github.com/padraic/mockery/blob/master/tests/Mockery/SpyTest.php

What are still some undocumented parts of the code?

Hidden features that would fit nice into cookbook entries?

Add more known gotchas?

Do the docs need reorganizing? I know it's been done relatively recently, but was it completed, are most of the users happy with the new docs?

Would be nice to know the state of the docs from a "new user" point of view, I'm using mockery for... 3-4 years now, I kind of don't use the docs any more.

robertbasic:
These are some good issues to move over here https://github.com/padraic/mockery-docs/issues

padraic:
We can get a list going somewhere under the new label perhaps...

davedevelopment:
Top of my head....

Some pages don't have code samples, we should probably put some in even for the simplest things, e.g. http://docs.mockery.io/en/latest/reference/public_properties.html

I think some of the headings could be clearer.

godinaa:
I recommend reordering the way in which examples are presented. Right now it reads as example 1 code, example 1 explanation, example 2 code, example 2 explanation, etc. Rather than describing the code above it, it should describe what it's about to show you.

I though this was a conventional technical writing practice, but it may just be a personal preference.

knugen:
I am of the exact same opinion as @aarongodin. It annoyed me enough to dig this old issue up :)

I would personally prefer it if the order was flipped, but if the maintainers for any reason want to stick to this ordering it would help if each example/explanation pair were visually grouped together.

If a decision is made on this I wouldn't mind helping out with the PR.

davedevelopment:
Regarding the order, I think I would prefer 

> explanation
> code
> further explanation

Regarding the whole of this issue, I'm currently considering rewriting the docs from scratch. Some of the pages we have cover things that aren't completely necessary, some are convoluted etc.

I started this morning, trying to get a feel for how I might write them. Still not settled on an outline or anything like that, but people are welcome to take a look at https://github.com/davedevelopment/mockery/tree/new-docs.

knugen:
I think that would be a very good structure. Thumbs up!

robertbasic:
Uhhh... I guess this is still waiting for me.

padraic:
[...]

PR #735, opened 2017-05-09
Add explicit pattern matcher

### LINKED ISSUE #338: Expectation argument matching thinks json strings are regexes
I had a problem with expecting a json string as an argument today. It would match no matter what, because it thinks '{}' are delimiters of a regex.

Example:
expected: {"tempVar1":"content","tempVar2":"content","tempVar3":"content"}
actual: {"tempVar1":"content","tempVar2":"content","tempVar3":"content","muh":123}

matches just fine.

I know i could force a string comparison by using a hamcrest matcher, but the above was kinda unexpected.

robertbasic:
Stop trying to do implicit regex matching when matching
string arguments, as it tends to not play nicely with
JSON strings as arguments - PHP thinks it's a regex pattern
to match against.

Introduce \Mockery::pattern() and the Pattern matcher to do
explicit regex pattern matching when it's required by the user.

Fixes #338

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 * Marked `Mockery\MockInterface` as internal
 * Subset matcher matches recusively
 * BC BREAK - Spies return `null` by default from ignored (non-mocked) methods with nullable return type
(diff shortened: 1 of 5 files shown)


robertbasic:
The documentation for this is going to be updated in #442

PR #736, opened 2017-05-09
Track internally bad method calls

### LINKED ISSUE #613: Mockery exceptions being caught in tested code
As per title. When an expectation fails on a Mockery mock, it throws an exception which is caught by application exception catchers instead of being aggregated at the test case level. It's impossible to know what expectation failed unless you actually go in code to where the exception is being caught and dump the exception for information.

robertbasic:
Instead of throwing bad method call exceptions, we now throw them,
catch them, add them to an internal array of bad method calls, and
then when verifying the mock itself, we rethrow these exceptions
again.

This way we work around for user application code that does a
try/catch block on the general \Exception, effectively catching
Mockery's exceptions and silencing them.

Fixes #613

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,10 @@ matrix:
     - php: 5.6
       env:
         - DEPS=latest
+    - php: 5.6
+      env:
+        - PHPUNIT=minimum
+        - DEPS=latest
     - php: 7.0
       env:
         - DEPS=lowest
@@ -47,6 +51,7 @@ before_install:
   - alias composer=composer\ -n && composer self-update
 
 install:
+  - if [[ $PHPUNIT == 'minimum' ]]; then sed -i 's/~5.7|/5.4.*|/g' ./composer.json ; fi
   - if [[ $DEPS == 'latest' ]]; then travis_retry composer update --no-interaction ; fi
   - if [[ $DEPS == 'lowest' ]]; then travis_retry composer update --prefer-lowest --prefer-stable --no-interaction ; fi
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,9 @@
 * Marked `Mockery\MockInterface` as internal
 * Subset matcher matches recusively
 * BC BREAK - Spies return `null` by default from ignored (non-mocked) methods with nullable return type
+* BC BREAK - Remove implicit regex matching when trying to match string arguments, introduce `\Mockery::pattern()` when regex matching is needed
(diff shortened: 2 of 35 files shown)


robertbasic:
I do think this is a bad idea, and application code should be fixed instead to not catch `\Exception`. 

Feel free to reject this PR with no explanations. Set it on fire. Throw it out the window.

davedevelopment:
My thoughts were we should still throw them, but record that we have done so.

```
$e = new BadMethod...
$this->thrownExceptions[] = $e;
throw $e;
```

I think then in the PHPUnit integration, if the test appears to have passed, we could interrogate each mocks `thrownExceptions` and try and act accordingly.

robertbasic:
OK, will make the changes.

robertbasic:
@davedevelopment I've made the changes, but now I like it even less then before. There's one failing test case I have no idea how to fix.

I don't think this should be in master. It's not worth it.

davedevelopment:
@robertbasic leave it here for now then. I'll take a look at some point at the failing test case at some point. 

This isn't a feature we desperately need. Perhaps we don't throw the exceptions but mark a test as risky if it has seemingly passed but we threw some exceptions, that could work.

davedevelopment:
@robertbasic See my additions in this branch https://github.com/davedevelopment/mockery/tree/robertbasic-fix/issue613

First is that if the user really is expecting to swallow one of our exceptions, they can tell us so

```
$e->dismiss();
```

The second part is we mark the test as risky, rather than failing it. I think this is slightly more BC and probably more friendly? Not sure.

robertbasic:
@davedevelopment much better than my approach!

Line 838 in Mock.php should use `$_mockery_thrownExceptions` and not my `$_mockery_badMethodCallExceptions`.

davedevelopment:
Fixed that bit. Do you want to pull my commits on to your branch to keep it all together in this PR, better than starting a fresh one. 

Testing this is hard, maybe we need to run a few adhoc tests that invoke PHPUnit from the outside.

robertbasic:
@davedevelopment I've fetch/merged your changes into my branch. I have no idea why some commits are duplicated. ¯\_(ツ)_/¯

davedevelopment:
Probably my fault...

Will pull this down and add some tests, then get this merged shortly.

luispabon:
This is good stuff, apps shouldn't have any knowledge of the testing framework / tooling being employed. Catching and handling \Exception at the very top of the app code path is paramount for correct error handling.

PR #751, opened 2017-05-22
Doc Fix invalid code for closure

fredjiles:
The condition was checking for `$arg` which is undefined and should have been checking for `argument`


davedevelopment:
Thanks!

Issue #757, opened 2017-06-06
Override parameter map for internal classes does not accept default values for parameters by reference


sGy1980de:
In case, you want to mock classes from internal modules, mockery will fail to create proper source code for methods with optional parameters passed by reference, e.g. http://php.net/manual/en/memcache.get.php

```php
\Mockery::getConfiguration()
     ->setInternalClassMethodParamMap(\Memcache::class, 'get', ['$id', '&$flags = null']);

// this will result in, which is invalid 
public function get($id,&$flags = null) {
    $argc = func_num_args();
    $argv = func_get_args();
    if ($argc > 1) {
        $argv[1] = &$flags = null;
    }
    $ret = $this->_mockery_handleMethodCall(__FUNCTION__, $argv);
    return $ret;
}
```

The generator needs to filter the default value for the for the $argv asssingment. The expected generated code would look like this.
```php
public function get($id,&$flags = null) {
    $argc = func_num_args();
    $argv = func_get_args();
    if ($argc > 1) {
        $argv[1] = &$flags;
    }
    $ret = $this->_mockery_handleMethodCall(__FUNCTION__, $argv);
    return $ret;
}
```

sGy1980de:
#758 will fix this

PR #758, opened 2017-06-06
Fix: internal class param by reference cannot have default value

sGy1980de:
This will fix issue: https://github.com/mockery/mockery/issues/757, which describes, that Mockery will actually fail to create proper code, when overriding the parameter map with optional parameters passed by reference, e.g. [Memcache::get()](http://php.net/manual/en/memcache.get.php).


davedevelopment:
Great catch and fix, thanks!

PR #759, opened 2017-06-07
support PHPStorm autocompletion

wan2land:
when I use mockery in phpstorm, mockery doesn't work autocompletion perfectly.

so i make `.phpstorm.meta.php`, how about apply this PR? :-)

**before**

**after**

### DIFF
--- a/.phpstorm.meta.php
+++ b/.phpstorm.meta.php
@@ -4,5 +4,14 @@ namespace PHPSTORM_META {
         \Mockery::mock('') => [
             "" == "@",
         ],
+        \Mockery::spy('') => [
+            "" == "@",
+        ],
+        \mock('') => [
+            "" == "@",
+        ],
+        \spy('') => [
+            "" == "@",
+        ],
     ];
 }
--- a/.phpstorm.meta.php
+++ b/.phpstorm.meta.php
@@ -1,0 +1,8 @@
+<?php
+namespace PHPSTORM_META {
+    $STATIC_METHOD_TYPES = [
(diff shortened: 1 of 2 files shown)


davedevelopment:
This looks interesting! I don't know much about PHPStorm, does it automatically scan `vendor/` for these meta files?

Are we able to do the same for the `Mockery::spy` method and also the `mock` and `spy` global functions? They're are the most commonly used ones.

wan2land:
yes. it also works in vendor directory.

i check the other method and global functions, then i will PR again. :-)

wan2land:
helper functions use `call_user_func_array`, but IDE(PHPStorm .. and so on...) cannot track it.

so, i change `call_user_func_array` to variadic function call. (mockery support php 5.6)


**before**



**after**

MarcusSchwarz:
Hey @wan2land , this is neat stuff! I did not know yet that this is possible in phpstorm. After reading the docs I noticed you are using a deprecated notation for the file, valid until version 2016.1
While there is no information on how long this old format will be supported, wouldn't it make sense to use the new format already? Or do you see any show stoppers there?

@davedevelopment, just out of curiosity, which editor are you using?

wan2land:
i don't know new format. i will check new format!
so, close this PR. :-)

davedevelopment:
@wan2land thank you for the work so far and look forward to the new PR!

@MarcusSchwarz I use vim, but I'm not some pretentious dude who thinks he's awesome for it, just stuck in my old ways!

Issue #760, opened 2017-06-16
ExpectationInterface seems incorrect with implementations 


gsdevme:
Hi,

I keep noticing this within Unit tests that my IDE is flagging up a mismatch between the implementation and interface for `ExpectationInterface` 

Is there a reason `ExpectationInterface` has zero arguments on its `andReturn()`?

PR #761, opened 2017-06-16
chore: improve docblocks / types for a better experience

gsdevme:
Fixes https://github.com/mockery/mockery/issues/760

Before
screen shot 2017-06-16 at 12 47 56

After
screen shot 2017-06-16 at 12 43 12


davedevelopment:
Thanks!

PR #766, opened 2017-07-09
Custom object formatters

MartinMystikJonas:
Possibility to set custom objet formatters for specific classes or interfaces. If formatter is set it's used instead of default formatter outputing public properties.

```
        \Mockery::getConfiguration()->setObjectFormatter(
            'ClassWithCustomFormatter',
            function($object) {
                return array(
                    "properties" => array(
                        "stringProperty" => $object->stringProperty
                    ),
                    "getters" => array(
                        "gettedProperty" => $object->getArrayProperty()
                    )
                );
            }
        );
```

### DIFF
--- a/library/Mockery.php
+++ b/library/Mockery.php
@@ -662,10 +662,25 @@ class Mockery
             return array('...');
         }
 
-        return array(
-            'class' => get_class($object),
-            'properties' => self::extractInstancePublicProperties($object, $nesting)
+        $class = get_class($object);
+        $customFormatterClass = self::getConfiguration()->findObjectFormatter($class);
+
+        $array = array(
+          'class' => $class,
+          'identity' => '#' . md5(spl_object_hash($object))
         );
+
+        if ($customFormatterClass) {
+            $customFormatter = self::getConfiguration()->getObjectFormatter($customFormatterClass);
+            $array = array_merge($array, $customFormatter($object));
+        } else {
+            $array = array_merge(
+                $array,
+                array('properties' => self::extractInstancePublicProperties($object, $nesting))
+            );
+        }
+
+        return $array;
     }
 
     /**
@@ -685,7 +700,11 @@ class Mockery
         foreach ($properties as $publicProperty) {
             if (!$publicProperty->isStatic()) {
                 $name = $publicProperty->getName();
-                $cleanedProperties[$name] = self::cleanupNesting($object->$name, $nesting);
+                try {
+                    $cleanedProperties[$name] = self::cleanupNesting($object->$name, $nesting);
+                } catch (\Exception $exception) {
+                    $cleanedProperties[$name] = $exception->getMessage();
+                }
             }
         }
 
--- a/library/Mockery.php
+++ b/library/Mockery.php
@@ -662,23 +662,20 @@ class Mockery
             return array('...');
         }
 
+        $defaultFormatter = function ($object, $nesting) {
+            return array('properties' => self::extractInstancePublicProperties($object, $nesting));
+        };
+
         $class = get_class($object);
-        $customFormatterClass = self::getConfiguration()->findObjectFormatter($class);
+
+        $formatter = self::getConfiguration()->getObjectFormatter($class, $defaultFormatter);
 
         $array = array(
           'class' => $class,
           'identity' => '#' . md5(spl_object_hash($object))
         );
 
-        if ($customFormatterClass) {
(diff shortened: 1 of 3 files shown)


library/Mockery/Configuration.php
```diff
@@ -127,4 +134,27 @@ public function getInternalClassMethodParamMaps()
...
+    {
+        $parentClass = $class;
+        for ($classes[] = $parentClass; $parentClass = get_parent_class($parentClass); $classes[] = $parentClass);
+        $classesAndInterfaces = array_merge($classes, class_implements($class));
+        foreach ($classesAndInterfaces as $type) {
+            if (isset($this->_objectFormatters[$type])) {
+                return $type;
+            }
+        }
+        return null;
```
davedevelopment: If we return the default here, we can remove the if/else [when calling](https://github.com/mockery/mockery/pull/766/files#diff-6326a23e84d7f7c0f0b92d031ba68bd7R594)?

```
return function ($object, $nesting) {
    return ['properties' => self::extractInstancePublicProperties($object, $nesting)];
};
```

library/Mockery/Configuration.php
```diff
@@ -127,4 +134,27 @@ public function getInternalClassMethodParamMaps()
...
+
+    public function setObjectFormatter($class, $formatterCallback)
+    {
+        $this->_objectFormatters[$class] = $formatterCallback;
+    }
+
+    public function findObjectFormatter($class)
+    {
+        $parentClass = $class;
+        for ($classes[] = $parentClass; $parentClass = get_parent_class($parentClass); $classes[] = $parentClass);
```
robertbasic: Would you mind rewriting this for loop so it is easier to understand? Thanks!

robertbasic:
@MartinMystikJonas hey, thanks for the PR! Can you fix the conflicts and update the PR based on @davedevelopment's and mine requests? Thank you!

library/Mockery/Configuration.php
MartinMystikJonas: It is not possible because extractInstancePublicProperties is private

library/Mockery/Configuration.php
MartinMystikJonas: Rewritten

MartinMystikJonas:
Finally had time to finish this. Sorry about three years gap but after our daughter was born I had only little time for open source :-)

library/Mockery/Configuration.php
MartinMystikJonas: Maybe pass default resolver as paramater?

MartinMystikJonas: Refactored in following commit https://github.com/mockery/mockery/pull/766/commits/10cf1d7c9afeb3dd5fd22f86e46339d309880b42

MartinMystikJonas:
Let me know if you want to merge it I will squash commits

MichaelKubovic:
Friendly ping, is there anything we can do to merge this?

library/Mockery.php
```diff
@@ -662,10 +662,22 @@ private static function objectToArray($object, $nesting = 3)
...
+            return array('properties' => self::extractInstancePublicProperties($object, $nesting));
+        };
+
+        $class = get_class($object);
+
+        $formatter = self::getConfiguration()->getObjectFormatter($class, $defaultFormatter);
+
+        $array = array(
+          'class' => $class,
+          'identity' => '#' . md5(spl_object_hash($object))
```
GrahamCampbell, line 675: why?

GrahamCampbell (changes requested):

library/Mockery.php
MartinMystikJonas: It is often useful to be able to compare object identities not only their class + some optional custom info from formatter. Should I remove it?

GrahamCampbell: I mean why the md5. Just use `spl_object_id`.

MartinMystikJonas: Besause spl_object_id is hard to compare visually becaue usually two instances have 95% characters of spl_object_hash identical. Using of md5 of spl_object_hash is quite common practice i saw in many libraries.

MartinMystikJonas: For example I just used it on two objects and I got:
000000002f73941b000000004f808351 and 000000002ffa4e380000000026e9a821
If you use md5 it becomes different at the first sight:
69850868ab17d5cd4d7effccd55ba63c and 43147e68957af1fcbd045979c3c3fda3

Issue #742, opened 2017-05-13
Risky Tests


davedevelopment:
Fix them...

robertbasic:
This is on PHP 7.1 only, right?

padraic:
Wherever you find PHPUnit 6, so also PHP 7.0.

PR #769, opened 2017-07-24
Fix risky tests

MarijnKoesen:
Using mockery 1.0-dev we get a lot of warnings and [risky tests](https://github.com/mockery/mockery/issues/742) this PR fixes the issue.

In Mockery.php on line 163 there's an `@unlink` which phpunit warns about if xdebug is enabled.

```
unlink() used in /Users/marijn/my-project/vendor/mockery/mockery/library/Mockery.php:163
```

To fix this you can add Mockery to the Blacklist in phpunit so it's ignored and tests are no longer marked risky because of this `@unlink`.

--

Maybe there is a better way/place to put this, as I read that the Listener is kind of deprecated in version 6.  But I couldn't find a better place. As the Listener allows you to do it `onSuiteStart` which is better than always doing it `onTestStart`.

The Listener also makes the user aware in case you don't call `Mockery::close`, which is nice, as we've already had some places where we didn't call it (because we forgot to use the Trait), and the tests where green, while in fact the expectations were not met.

So I actually like to use the Listener for this sake. Although the downside is that you need every test to use the trait, even when you don't use expectations. But I guess that's a trade-off unless that can be fixed.

Let me know if there is a better place and it should be moved to somewhere else.

/edit: I noticed the tests are failing with old phpunit version still, unfortunately. But let's discuss the solution we want to go for first, then i'll fix it properly. (either another/new listener, or a listener-less way)

### DIFF
--- a/library/Mockery/Adapter/Phpunit/TestListener.php
+++ b/library/Mockery/Adapter/Phpunit/TestListener.php
@@ -69,4 +69,11 @@ class TestListener extends BaseTestListener
         $result = $test->getTestResultObject();
         $result->addFailure($test, $e, $time);
     }
+
+    public function startTestSuite(\PHPUnit\Framework\TestSuite $suite)
+    {
+        \PHPUnit\Util\Blacklist::$blacklistedClassNames[\Mockery\Mockery::class] = 1;
+
+        parent::startTestSuite($suite);
+    }
 }
--- a/tests/Mockery/Adapter/Phpunit/TestListenerTest.php
+++ b/tests/Mockery/Adapter/Phpunit/TestListenerTest.php
@@ -83,6 +83,15 @@ class Mockery_Adapter_Phpunit_TestListenerTest extends TestCase
         $mock->bar();
         \Mockery::close();
     }
+
+    public function testMockeryIsAddedToBlacklist()
+    {
+        $suite = \Mockery::mock(\PHPUnit\Framework\TestSuite::class);
+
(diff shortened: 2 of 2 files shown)


davedevelopment:
@MarijnKoesen thank you for this. I decided to drop support PHPUnit 5.x in the TestListener, see #786. 

I'm happy with this going in the listener for now, so if you want to update this PR (or create a new one) with just the second commit, I'll see it merged.

MarijnKoesen:
Thanks, i'll update the PR, but am on holiday right now, I'll update it when I'm back, in about two weeks :+1:

MarijnKoesen:
@davedevelopment  The PR is updated, tests are green now. 

I've also created another PR that fixes the codestyle (in master): https://github.com/mockery/mockery/pull/811

MarijnKoesen:
From this (lots of risky, and also 156% progress? wtf?):

screen shot 2017-11-22 at 16 23 56

To no risky tests and 100% progress:

screen shot 2017-11-22 at 16 32 23


This fixes #742

rafaelbeckel:
Still having the same issue. I had to migrate my mocks to native PHPUnit syntax to make it count the assertions.

I will build a Docker container with the bug in a few days, so you can reproduce it.

MarijnKoesen:
What are the warnings you are getting?

We are getting them again as well, now on `vendor/guzzlehttp/psr7/src/functions.php:81`, but this is not a class, so I don't think we can add them to the blacklist, like we did with the Mockery class ([here](https://github.com/mockery/mockery/blob/master/library/Mockery/Adapter/Phpunit/TestListener.php)... 🤔

PR #771, opened 2017-07-26
Cast expectedArgs to array (fix php7.2)

fernandobandeira:
On PHP 7.2 count throws a warning when passing something that isn't an array or a implementation of Countable to the count function.

``` _expectedArgs ``` may be null sometimes, this causes [tests to fail](https://travis-ci.org/laravel/framework/jobs/257748808) on 7.2.


duncan3dc:
Casting to arrays is one of the few things in the PHP language that is undefined:
_"The behaviour of an automatic conversion to array is currently undefined"_
http://php.net/manual/en/language.types.type-juggling.php

I tackled this same problem in #718

Issue #264, opened 2014-02-02
Allow setting exception message for `InvalidCountException` exception


aik099:
I like use `->once()`, `->twice()`, `->times()` and other methods that validate how much times a method should be called a lot.

The error I'm getting on failed expectation usually looks like this: `Method stop() from Mockery_2_Behat_Mink_Session should be called`. It looks good for a generic error, but it doesn't really explain what's wrong with calling code.

I'm proposing to develop a way to pass optional exception message to be used instead of generic one if an expectation fails. Like with `->assertEquals(..., ..., $message)` - you can pass a message if you wish, but when not passed the generic is used.

padraic:
Anyone have any additional thoughts on this? If someone throws a PR at it, I wouldn't be against either API option.

ianchadwick:
Something like this?
 https://github.com/ianchadwick/mockery/commit/9b8d4a3c0b96aef4a83086d07e39beb5c286d5c6

Seemed pretty easy to implement which from the above discussion makes me think maybe I'm missing something...

davedevelopment:
@ianchadwick that looks alright to me, open a PR and we can discuss from there :+1:

PR #778, opened 2017-08-14
Set the InvalidCountException message using because($message);

ianchadwick:
Adds functionality for #264 

```php
// Add a message which will be displayed in the exception message
$mock->shouldReceive('foo')->once()->because('We like foo');
```

### DIFF
--- a/library/Mockery/CountValidator/Exact.php
+++ b/library/Mockery/CountValidator/Exact.php
@@ -33,12 +33,15 @@ class Exact extends CountValidatorAbstract
     public function validate($n)
     {
         if ($this->_limit !== $n) {
+            $because = $this->_expectation->getExceptionMessage();
+
             $exception = new Mockery\Exception\InvalidCountException(
                 'Method ' . (string) $this->_expectation
                 . ' from ' . $this->_expectation->getMock()->mockery_getName()
                 . ' should be called' . PHP_EOL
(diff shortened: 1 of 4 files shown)


library/Mockery/CountValidator/Exact.php
```diff
@@ -33,12 +33,15 @@ class Exact extends CountValidatorAbstract
...
         if ($this->_limit !== $n) {
+            $because = $this->_expectation->getExceptionMessage();
+
             $exception = new Mockery\Exception\InvalidCountException(
                 'Method ' . (string) $this->_expectation
                 . ' from ' . $this->_expectation->getMock()->mockery_getName()
                 . ' should be called' . PHP_EOL
                 . ' exactly ' . $this->_limit . ' times but called ' . $n
                 . ' times.'
+                . ($because ? ' Because ' . $this->_expectation->getExceptionMessage() : '')
```
davedevelopment: Do you think we should change the `Because` to `Reason:`?

> should be called exactly 1 times but called 0 times. Reason: We like foo

ianchadwick: Could do :) This is consistent with the method call but Reason makes sense also, I'll let you call it...

davedevelopment:
@ianchadwick I decided that whole exception message needs reviewing anyway, so it can be considered then! Thanks for the PR.

PR #789, opened 2017-09-20
allows does not allow multiple return values

### LINKED ISSUE #788: error in docs for allows()
```
$mock = \Mockery::mock('MyClass');
$mock->allows([
    'name_of_method_1' => ['1st call return value', '2nd call return value'],
    'name_of_method_2' => ['1st call return value', '2nd call return value'],
]);
```

This isn't correct. The whole of ``` ['1st call return value', '2nd call return value']``` is returned from the call.

robertbasic:
Allows apparently does not allow to define multiple
return values for subsequent calls to the same method.

Not sure where I got this from the first time around I
wrote these docs?

Fixes #788

### DIFF
--- a/docs/reference/alternative_should_receive_syntax.rst
+++ b/docs/reference/alternative_should_receive_syntax.rst
@@ -20,8 +20,8 @@ all, were they called.
 
     $mock = \Mockery::mock('MyClass');
     $mock->allows([
-        'name_of_method_1' => ['1st call return value', '2nd call return value'],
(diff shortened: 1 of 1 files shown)


davedevelopment:
I don't know either, I don't think it was ever in there....

davedevelopment:
@robertbasic Just a quick note, I've turned those coveralls comments off, they do my head in. I've set it to fail a build if coverage drops by 5% as a fail safe.

robertbasic:
@davedevelopment oh yes please and thank you!

PR #794, opened 2017-10-05
Allows use of string method names in allows and expects

davedevelopment:
What do we think?

I personally prefer to avoid the string names, but this makes the API more like the old API, at the slight cost of confusing return types for the `allows` and `expects` methods, although the `allows` already has mixed return thanks to the array of stubs functionality.

### DIFF
--- a/tests/Mockery/AllowsExpectsSyntaxTest.php
+++ b/tests/Mockery/AllowsExpectsSyntaxTest.php
@@ -67,10 +67,27 @@ class AllowsExpectsSyntaxTest extends TestCase
     }
 
     /** @test */
+    public function allowsCanTakeAString()
+    {
+        $stub = m::mock();
+        $stub->allows("foo")->andReturns("bar");
+        $this->assertEquals("bar", $stub->foo());
+    }
+
+    /** @test */
     public function expects_can_optionally_match_on_any_arguments()
     {
         $mock = m::mock();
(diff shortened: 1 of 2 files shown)


tests/Mockery/AllowsExpectsSyntaxTest.php
```diff
@@ -76,6 +84,15 @@ public function expects_can_optionally_match_on_any_arguments()
     }
 
     /** @test */
+    public function expects_can_take_a_string()
+    {
+        $mock = m::mock();
+        $mock->allows("foo")->andReturns(123);
```
robertbasic: This tests `allows` and not `expects`.

robertbasic:
I don't really see the benefit of this? Why would this be needed/required?

davedevelopment:
@robertbasic it's not *needed*, but I actually find myself doing it, and then wondering why it doesn't work.

I also think it's better when you don't care what arguments come:

``` php
// old syntax
$mock->shouldReceive("foo")->atLeast()->once();

// new syntax as it is (no string method names, but have to specify any args)
$mock->expects()->foo(anyArgs());

// new syntax with this patch, string method name, but anyArgs avoided
$mock->expects("foo");
```

robertbasic:
@davedevelopment good enough reason for me. Fix the test and it's +1 from me.

PR #799, opened 2017-10-06
Finalises allows and expects syntax in API

### LINKED ISSUE #793: Pull up `expects()` method to the `MockInterface`
Hi!

I've always used the following method in order to set expectations:
```php
$repository = Mockery::mock(SomeRepository::class);
$repository->shouldReceive('search')->once()->andReturn($xxx);
```

I'm using `"mockery/mockery": "^1.0@dev"` for the first time in order to be able to upgrade to `"phpunit/phpunit": "^6.3"`. I've seen in [the Method Call Expectations ReadMe section](https://github.com/mockery/mockery/blob/master/README.md#method-call-expectations-) that you recommend the usage of the `expects()` method. It seems a little bit clearer for me setting the expectations this way, so I've bet for it 🙂:
```php
$repository = Mockery::mock(SomeRepository::class);
$repository->expects()->search()->once()->andReturn($xxx);
```

The problem comes when I try to type hint this `$repository` instance. If I type hint it to a `\Mockery\Mock` instance, I receive the following error while running the tests:

>TypeError: Argument 1 passed to someMethod() must be an instance of Mockery\Mock, instance of Mockery_0_All_My_Whole_Namespace_SomeRepository given

So in order to make it run without type errors, I have to type hint to the `\Mockery\MockInterface` interface. However, the `expect()` method signature is not defined in this interface, so calling to this method while type hinting to the interface ends up believing in something more related to Duck Typing and runtime faith, rather than the actual type system.

**What do you think about pulling up the `expects()` method signature to the `\Mockery\MockInterface` interface in order to be able to properly use type hint?**

davedevelopment:
Fixes #793


robertbasic:
Looks good to me! :+1:

davedevelopment:
@letr0n sadly this was a  BC break for 1.0

PR #808, opened 2017-11-08
Deprecate allowMockingMethodsUnnecessarily

### LINKED ISSUE #804: allowMockingMethodsUnnecessarily method is not working
I am new to Mockery and when I came out on documentation page about Mockery global configuration, I read about **allowMockingMethodsUnnecessarily** function, which purpose is to notice/don't notice about not fulfilled expectations.  I added it to a part in my code which is included everywhere, but I saw no changes applied. 
My code for setting Mockery global configuration:
`\Mockery::getConfiguration()->allowMockingMethodsUnnecessarily(false);`

My unit test part which is not telling me about mistake
` $myClassMock = \Mockery::mock('alias:' . MyClass::class);`
`$myClassMock->shouldReceive('isValueValid')->once()->andReturn(true);`
`$myClassMock->shouldReceive('createErrorMessage');`

The code part unit test is written for.

> 
        $validationErrors = [];
        $isValid = MyClass::isValueValid(
            'Valid value',
            [
                new NotEmpty(),
            ],
            $validationErrors
        );

        if (!$isValid) {
            $message = MyClass::createErrorMessage($validationErrors)
             
           var_dump($message);
        }

>         

From this code part you can guess that, the units must complain about 
`$myClassMock->shouldReceive('createErrorMessage');`
Because `createErrorMessage` is never called. But I get nothing.

When I put change this line like `$myClassMock->shouldReceive('createErrorMessage')->once()`, unit test fails, but it is independent of value which is passed to `allowMockingMethodsUnnecessarily` function.

My question is, am I missing something when specifying global configuration or do I have some misunderstanding of the purpose of this config?

And also one little note: When I looked up in code I saw that `_allowMockingMethodsUnnecessarily` property and `mockingMethodsUnnecessarilyAllowed` method of `\Mockery\Configuration` class are never used from outside.

robertbasic:
Fixes #804

### DIFF
--- a/docs/mockery/configuration.rst
+++ b/docs/mockery/configuration.rst
@@ -9,32 +9,28 @@ configuration object to store a small subset of core behaviours. The three
 currently present include:
 
 * Option to allow/disallow the mocking of methods which do not actually exist
-* Option to allow/disallow the existence of expectations which are never
   fulfilled (i.e. unused)
 * Setter/Getter for added a parameter map for internal PHP class methods
   (``Reflection`` cannot detect these automatically)
 
-By default, the first two behaviours are enabled. Of course, there are
+By default, the first behaviour is enabled. Of course, there are
 situations where this can lead to unintended consequences. The mocking of
 non-existent methods may allow mocks based on real classes/objects to fall out
 of sync with the actual implementations, especially when some degree of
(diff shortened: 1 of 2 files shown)


davedevelopment:
Nice :+1:

PR #811, opened 2017-11-22
Fix codestyle to be compliant with the standards

MarijnKoesen:
The codestyle check was failing, this makes it green again.


robertbasic:
Thanks!

PR #814, opened 2017-12-01
Test against PHP 7.2

carusogabriel:
With the release of  [`PHP 7.2`](http://php.net/archive/2017.php#id2017-11-30-1), would be nice to test it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,6 +30,12 @@ matrix:
     - php: 7.1
       env:
         - DEPS=latest
+    - php: 7.2
+      env:
+        - DEPS=lowest
+    - php: 7.2
+      env:
+        - DEPS=latest
     - php: nightly
       env:
         - DEPS=lowest


davedevelopment:
Ah, I merged this thinking we'd be all good after #817. Perhaps we'll have to allow 7.2 failures with the lowest deps? 

I think the failure is related to this (non) bug https://bugs.php.net/bug.php?id=75590, which I've already had to deal with somewhere.

carusogabriel:
I thought this as well :disappointed: 

Do you allow it or may I submit a PR?

carusogabriel:
Actually, after reading the log, it is failing on PHP 7.2 with composer --prefer-lowest, because [it is installing `PHPUnit 5.7`](https://travis-ci.org/mockery/mockery/jobs/309957233#L515). I'm gonna search some fixes.

carusogabriel:
[I've commented](https://github.com/sebastianbergmann/phpunit/issues/2899#issuecomment-351968793) on a PHPUnit issue.

davedevelopment:
The each is deprecated, but we can get around the fatal by requiring 5.7.10 I think. Trying now.

davedevelopment:
#820

PR #818, opened 2017-12-14
Refactoring tests

carusogabriel:
I've refactored more tests, like #815, this time using:
- `assertInstanceOf` and `assertNotInstanceOf` instead of `instanceof` operator and `get_class` function;
- `assertArrayHasKey` and `assertArrayNotHasKey` instead of `array_key_exists` and `isset` function;
- `assertLessThan` for mathematical comparisons;
- `assertCount` instead of `count` function;
- `assertInternalType` instead of `is_callable` function.


davedevelopment:
Thanks again!

PR #821, opened 2017-12-15
Fix PHPdoc

xificurk:
Some methods contain phpdoc with return type of `ExpectationInterface`, but this interface contains only a very limited set of methods compared to the actual implementation in `Expectation`. This cases various problems with autocompletion in IDEs.
A previous attempt to solve this used `@method` annotations on `ExpectationInterface`, but this approach did not cover all the necessary methods, and furthermore is simply wrong - either every implementor supports all the methods and they should be defined in PHP code of the interface, or they don't in which case additional `@method` annotations are wrong.

### DIFF
--- a/library/Mockery/ExpectationInterface.php
+++ b/library/Mockery/ExpectationInterface.php
@@ -21,17 +21,6 @@
 
 namespace Mockery;
 
-/**
- * @method Expectation once()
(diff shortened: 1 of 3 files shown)


robertbasic:
Thanks @xificurk!

PR #826, opened 2018-01-03
Updated protected method mocking exception message

robbytaylor:
Sorry for the unsolicited PR.

I just had a test error from trying to mock a protected method. I thought it might be useful if the message explained how to enable mocking protected methods on the mock object. It seems like if a developer is getting the error message then there's a good chance they want to mock the protected method.


davedevelopment:
Thanks!

davedevelopment:
> Sorry for the unsolicited PR.

Don't ever be sorry for that!

robbytaylor:
Thanks!

PR #828, opened 2018-01-10
Fixes from PHPStan

carusogabriel:
Some fixes found with [`PHPStan`](https://github.com/phpstan/phpstan). Noticeable changes:
- `__construct` does not return anything;
- Fix [documentation of variadic parameters](https://github.com/phpstan/phpstan/issues/683#issuecomment-350254607);
- PHPUnit's `markAsRisky` [does not take any argument](https://github.com/sebastianbergmann/phpunit/blob/master/src/Framework/TestCase.php#L693).

### DIFF
--- a/library/Mockery.php
+++ b/library/Mockery.php
@@ -34,7 +34,7 @@ class Mockery
     /**
      * Global container to hold all mocks for the current unit test running.
      *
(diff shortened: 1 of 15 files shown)


davedevelopment:
Thanks!

Issue #822, opened 2017-12-15
Overloading of named mocks


agxmeister:
While covering the legacy code by tests I am faced with following challange: I have a hardcoded dependance on class and usage of constants of this class in the method I need to cover. It looks like the following:

```
public static function getTransport()
{
	if ((new Config())->get(Config::OPTION_ALLOW_RSYNC)) {
		...
	}
	...
}
```

I can workaround hardcoded dependance using "overload" feature like described here (http://docs.mockery.io/en/latest/cookbook/mocking_hard_dependencies.html?highlight=overload):

```
$configMock = \Mockery::mock('overload:Config');
$configMock->allows('get')->andReturn('expected');
```

Constants also could be workarounded using "namedMock" feature like described here (http://docs.mockery.io/en/latest/cookbook/class_constants.html?highlight=constant):

`$mock = \Mockery::namedMock('Config', 'ConfigStub');`

But how I can combine these two features together? In other words, how I can create a named mock and ensure that each newly created object of mocked class will support allowed methods?

robertbasic:
This is now possible since #829 with a map of constants to mock.

smitit173:
@robertbasic this is not possible if the constant is an array. It takes the array as a string. Is there any workaround for this?

PR #829, opened 2018-01-18
Map of constants to mock

robertbasic:
``` php
\Mockery::getConfiguration()->setConstantsMap([
    'ClassName' => [
        'CONSTANT_NAME' => 'CONSTANT_VALUE',
    ]
]);
```


robertbasic:
With this, #822 can be done by configuring a map of constants to mock, and then overloading that mocked class.

davedevelopment:
I don't love the usage, but it's probably the best we can do for now and should get some people on their way 👍

robertbasic:
@davedevelopment I don't like it either :/

Issue #831, opened 2018-01-23
Expectation#verify return value.


Johannestegner:
The `Expectation` class have a typehint on the `verify` method which is set to `bool`, while it does not return bool but rather nothing/void.  
Looking at other verify methods in the framework (E.G., `mockery_verify`), it should probably return `void` in typehint and just throw exceptions as it already do.

robertbasic:
Done with #832, thanks again @Johannestegner!

PR #832, opened 2018-01-23
Typehinted return value on Expectation#verify.

Johannestegner:
Changed Expectation#verify return value typehint from bool to void.
Issue reference: #831


robertbasic:
Thanks @Johannestegner!

PR #834, opened 2018-01-24
Pattern helper added.

dmitry-ivanov:
Another one helper function added.

### DIFF
--- a/library/helpers.php
+++ b/library/helpers.php
@@ -49,3 +49,10 @@ if (!function_exists("anyArgs")) {
         return new AnyArgs();
     }
 }
+
+if (!function_exists("pattern")) {
+    function pattern($expected)
+    {
+        return Mockery::pattern($expected);
+    }
+}


dmitry-ivanov:
PS: From the other hand, I can see now, that there are a lot of specific validation methods in Mockery: `any`, `type`, `ducktype`, `subset`, etc etc...

Probably it's not right, if only `pattern` would have it's own helper.

So, I think we should decide - do we want to provide such helpers at all or not?
Maybe, some prefixed functions for these needs like `mockery_pattern`, `mockery_subset` would be ok to prevent naming conflicts?

robertbasic:
If we add these helpers, there's a potential problem of clashing with Hamcrest and/or PHPUnit. I think we had lots of problems like that in the past. Prefixing them with `mockery_*` does indeed prevent that, but then I fail to see what's their advantage over just using `\Mockery::pattern` and similar.

dmitry-ivanov:
Yep, thanks for you reply. Then, I'll just close the issue.

Issue #531, opened 2016-01-08
Feature: expect subset of method parameters


Bilge:
Currently all or no method parameters must be expected. It should be possible to expect a subset of parameters. For example, if I am only interested in the first parameter I should be able to specify how the first matches but then be able to indicate to Mockery in some way that any additional parameters are unimportant, whether they're present or not.

aik099:
You can specify `m::any()` as 2nd parameter to allow matching for any 2nd+ argument values.

Bilge:
@aik099 I'm afraid that's simply not the case. `\Mockery::any()` will match exactly one argument in the position it appears in, only. That is, `->with(123, \Mockery::any())` is not the same as `->with(123, \Mockery::any(), \Mockery::any())`: the former will match exactly two arguments and the latter will match exactly three.

davedevelopment:
I use `andReturnUsing` to do this kind of thing, which is clunky, but I'm not really too sure what might be better.

Bilge:
I believe the only two potential options would be 

* a matcher that specifies the rest of the arguments are irrelevant, or
* a method that will only check the arguments specified and ignore the rest.

aik099:
> a matcher that specifies the rest of the arguments are irrelevant, or

I vote for this one. This how it was implemented in https://github.com/phpspec/prophecy :

* `AnyValueToken` or `Argument::any()` - matches any argument
* `AnyValuesToken` or `Argument::cetera()` - matches any arguments to the rest of the signature

davedevelopment:
And are we happy with `cetera`? I don't mind it, wondering if we could use `\Mockery::variadic()`?

Bilge:
I haven't used Prophecy but I admire it's inventiveness with respect to naming and the syntax of its API. That said, I don't think we should be lifting hip names like *cetera* straight from it because that's not Mockery's style. If we like that name, and it's certainly pretty good, then I think we should use the full name *etcetera* and/or the popular abbreviation, *etc*.

I'm sorry to say, but *variadic* is a poor choice because the matcher does not only apply to variadic functions and is more likely to be misleading.

Bilge:
I liked this issue better when it wasn't dead.

Bilge:
@davedevelopment @robertbasic Any further thoughts on this?

robertbasic:
@Bilge this isn't something I would personally use, but I'd accept a PR for it, if it provides a nice implementation.

robertbasic:
[...]

Issue #301, opened 2014-03-19
optional argument matcher


VladaHejda:
I want to suggest something like **OPTIONAL Matcher**, that will accept calling mocked method both with the optional argument and without it.

My code calls some method, sometimes with an optional argument, sometimes not:

``` php
$mock->method( $mandatoryArg, $optionalArg );
$mock->method( $mandatoryArg );
```

But for now, there is no way how to mock this by one specification, I have to do this:

``` php
$mock->shouldReceive( 'method' )
    ->with( 'mandatory', 'optional' );
$mock->shouldReceive( 'method' )
    ->with( 'mandatory' );
```

My suggestion is to mock it this way:

``` php
$mock->shouldReceive('method')
    ->with( 'mandatory', \Mockery::optional( 'optional' ) );
```

When optional parameter should be matched, this can work:

``` php
$mock->shouldReceive('method')
    ->with( 'mandatory', \Mockery::optional( \Mockery::any() ) );
```

_I found [similar issue](https://github.com/padraic/mockery/issues/167), but it requires slightly different behavior._

padraic:
Closing issue as for matching. Should use Hamcrest and PR there if missing a specific need.

aik099:
Unfortunately this is no ordinary matcher, because when mocking methods with optional arguments Mockery knows their default value (from reflection) and for that reason we can assume every `->with('madatoryValue`)` call to be equivalent to `->with('madatoryValue', `optionalValue`)`. Without this users have to pass `m::any()` for every optional parameter.

padraic:
@aik099 Point taken. Reopening for now.

robertbasic:
@padraic @davedevelopment is this still something we'd consider adding?

tflori:
@aik099 I undertsand correct that `->shouldReceive('call')->with(m::type('string'), m::any())` matches `->call('something')` and `->call('anything', 'else')`?

davedevelopment:
@robertbasic I'm on the fence, it's a very narrow use case.

aik099:
`->shouldReceive('call')->with(m::type('string'), m::etcetra())` should match both of your calls.

tflori:
@aik099 I don't know etcetera? where to find it?

aik099:
[...]

PR #840, opened 2018-02-02
Adds withVariadicArgs for optional argument matching

robertbasic:
I realised that #531 and #301 can be done with `withArgs` and a closure, so adding `withVariadicArgs` as a shortcut for that.

Thoughts?

### DIFF
--- a/library/Mockery/Expectation.php
+++ b/library/Mockery/Expectation.php
@@ -433,6 +433,39 @@ class Expectation implements ExpectationInterface
     }
 
     /**
+     * Expect a variable number of arguments
+     *
+     * Useful when we want to set up an expectation only for a subset of arguments:
+     *
+     * ``` php
+     * class Foo
+     * {
+     *   public function bar($x, $y) {}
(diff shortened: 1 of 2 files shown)


robertbasic:
I'm not even entirely sold on having this in Mockery. I'd never use this, I'll stay with typing out `m::any()` for any argument I don't care about. Why?

If `foo($care, $dontCare)` gets changed to `foo($care, $alsoCare, $dontCare)` this optional argument matching will "swallow" that change, i.e. the tests won't catch that there's suddenly a 2nd param we care about. Otherwise the tests will break with "hey you passed in 2 arguments, but you should pass in 3".

To me, having it as a separate `with*` method makes it stand out more. Name of the method can be changed.

We can go with a matcher, or with a new method, but in either case, a big fat warning in the docs about the downsides of using it should be there.

Or yea, we can not have this at all :)

robertbasic:
Mulling things over, I think this should be a new matcher after all.

Issue #249, opened 2014-01-17
Assertion count should not include stubs


davedevelopment:
Currently, stub expectations (zeroOrMoreTimes) are counted as assertions when using the PHPUnit plugin, I'd vote to exclude them and only count actual count expectations.

Thoughts?

padraic:
Just checking - we all agreed to implement or has it been implemented?

aik099:
This was imlemented in https://github.com/padraic/mockery/blob/master/library/Mockery/Adapter/Phpunit/MockeryTestCase.php because due PHPUnit test listener limitation assertion count updater code had no effect.

davedevelopment:
@aik099 I don't think it was. This is specifically about not counting stubbed methods and only counting actual mock expectations. So

```
$first->shouldReceive("dave")->andReturn("foo");
$first->shouldReceive("dave1")->andReturn("foo");
$first->shouldReceive("dave2")->andReturn("foo");
$second->shouldReceive("doSomething")->once();
```

Would record one assertion.

@padraic I think we've all agreed, but not been implemented. Will try and get on to it this week.

robertbasic:
@davedevelopment are we still interested in this? If so, I think I can make a PR for it.

robertbasic:
Closing as there wasn't much feedback on it for quite some time.

PR #844, opened 2018-02-04
Include only call count constrained expectations in expectation count

robertbasic:
An attempt at #249, not sure if it's the right approach.

The `once()` calls in tests are to make them not be risky.


davedevelopment:
Nice!

Do you think we need to hold on this one until 2.0 though? Are we going to get loads of people complaining that we broke their suite with loads of risky tests?

robertbasic:
Relabeled the ticket towards 2.0, so I'm ok with that!

ghostwriter:
Resolved via #1180

PR #849, opened 2018-02-09
Test with docker

robertbasic:
This is not required to be accepted, I mainly made this for myself, to make my job easier.

Run `make test` and it'll run the tests in a docker container. By default it'll be PHP 7.2, but the PHP version can be changed with `export MOCKERY_PHP_VER=mockery_php71`.

When running for the first time for a specific PHP version, it'll pull/build the docker images required for it to run, every next time it'll reuse the images. So, if you run it only for PHP 7.0, only the images for PHP 7.0 will be pulled/built.

For PHP 5.6 there's a custom image being built, so that the composer deps get installed.

`make help` spits out a short help.


robertbasic:
BTW, it's meant to be a test environment only, so xdebug for code-coverage, ext/mongo, phpredis, are intentionally missing, to make the images as small as possible.

davedevelopment:
Big :+1: for a Makefile, but I'd prefer to keep `make test` as light as possible and I'd use it for quick test runs, e.g.

``` Make
.PHONY: test deps

vendor/composer/installed.json:
    composer install

deps: vendor/composer/installed.json

test: deps
    php vendor/bin/phpunit
```

Generally I'm +1 for the docker, but I just wonder if it would be simpler to forego the docker-compose file and have the docker commands directly in the make file? 

``` Make

test-all: test-7.2 test-7.1 test-7.0 

test-7.2: deps
    docker run -it --rm -v "$$PWD":/opt/mockery -w /opt/mockery php:7.2-cli php vendor/bin/phpunit

test-7.1: deps
    docker run -it --rm -v "$$PWD":/opt/mockery -w /opt/mockery php:7.1-cli php vendor/bin/phpunit

test-7.0: deps
    docker run -it --rm -v "$$PWD":/opt/mockery -w /opt/mockery php:7.0-cli php vendor/bin/phpunit

# whatever the command might be for 5.6 etc

```

I guess that removes a little flexibility :thinking:

robertbasic:
@davedevelopment updated to remove the dependency on docker-compose.

Makefile
```diff
@@ -0,0 +1,34 @@
+.PHONY: test deps build56
```
davedevelopment: Needs `test-all`, `test-all-7` etc?

robertbasic:
@davedevelopment works the same, doesn't require docker-compose, so I guess it's better like this.

Makefile
robertbasic: Not sure to be honest, my first time writing a Makefile :)

PR #851, opened 2018-02-14
Explicitely marks each target as PHONY

lucasvanlierop:
This way it's more obvious that a target will no result in creating a
file or directory also a single list tends to get out of date


davedevelopment:
I did not know you could do that! Cool, thank you.

PR #852, opened 2018-02-14
Vendor should be updated if composer.json changes

lucasvanlierop:
Makes `composer.json` a pre requisite of vendor, this is way `vendor` will be updated when `composer.json` is updated (useful if you run it locally)


robertbasic:
Thanks @lucasvanlierop!

PR #854, opened 2018-02-21
Adds support for late static binding

robertbasic:
Objects that are mocked and then "recreated" with
late static binding would not inherit the expectations set
on the original mock object.

By telling Mockery to consider those mock objects as instance
mocks as well, they will inherit the expectations as expected.

Went with a new prefix of `lsb:` instead of trying to bolt it
on top of existing functionality because I'm too scared of the
things that might break otherwise.

### DIFF
--- a/library/Mockery/Container.php
+++ b/library/Mockery/Container.php
@@ -141,6 +141,13 @@ class Container
                 $builder->addTarget('stdClass');
                 $builder->setName($name);
                 continue;
+            } elseif (is_string($arg) && substr($arg, 0, 4) == 'lsb:') {
+                $name = str_replace('lsb:', '', $arg);
+                $builder->setInstanceMock(true);
+                $builder->addTarget($name);
+                array_shift($args);
+
+                continue;
             } elseif (is_string($arg) && substr($arg, strlen($arg)-1, 1) == ']') {
                 $parts = explode('[', $arg);
                 if (!class_exists($parts[0], true) && !interface_exists($parts[0], true)) {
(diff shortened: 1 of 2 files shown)


davedevelopment:
I don't really understand the use case for this, the original report was back in 2013 and nobody really provided a concrete real life example.

I'm also not convinced the naming is right, or maybe I don't want to hijack `lsb` for something that is actually about something else. Perhaps mixing `instance` in to the name as well, or perhaps even just instance, as that's what we're doing? We're not creating a late static bound mock, we're creating an instance mock, and PHP is doing the late static binding?

davedevelopment:
Just to be clear:

``` php
$mock = instanceMock(Foo::class);
```

Would load Foo::class and mock it as normal, but $mock would contain the master set of expectations etc?

One problem with that is we already have a `Mockery::instanceMock` method, that seems to do the same as `Mockery::mock` 👀 

I'm not even sure `instance` mock makes total sense as a name?

robertbasic:
As for the use case, if a mocked object gets `new static()`-ed, the expectations get lost on the new object, even though it still implements the `MockInterface`.

davedevelopment:
@robertbasic sure, I understand what it does, I just meant I'm not sure what sort of procedure I'd be testing when I would want this.

robertbasic:
Ah, that... Yea, for that I'm not sure either.

I was probably more focusing on "could I" when doing this PR, instead of "should I".

robertbasic:
Closing as there's really no reason to have this in Mockery.

PR #857, opened 2018-02-26
Publish API Documentation to GH Pages

robertmain:
Automatically publish php docs to gh-pages. For this to work you will need to able gh-pages and configure it for the `gh-pages` branch.

You will also need to add an environment var to travis called `GITHUB_TOKEN` with your GitHub token (see https://docs.travis-ci.com/user/deployment/pages/). 

Finally, this does not provide a fully comprehensive overview of the API(there are docblocks missing, and phpdocumentor will yell about this in the "Errors" report with ~500 notices). But over time I'm sure that could be PR'd back in separately.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ phpunit.xml
 *.DS_store
 .idea/*
 .php_cs.cache
+apidocs/
--- a/.travis.yml
+++ b/.travis.yml
@@ -86,3 +86,14 @@ notifications:
     - dave@atstsolutions.co.uk
 
   irc: irc.freenode.org#mockery
+deploy:
+  overwrite: true
+  provider: pages
+  file_glob: true
+  file: apidocs/*
+  local_dir: apidocs
+  skip_cleanup: true
+  github_token: $GITHUB_TOKEN
+  on:
+    branch: master
+    php: '7.1'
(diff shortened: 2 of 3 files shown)


robertbasic:
It would be good if the API docs could be in the `docs/` folder somehow with the rest of the documentation, so that it eventually all lives in the same place.

robertmain:
Agreed. How do you want to proceed with that? I could make it so that it puts the API docs in `docs/api` and then it just deploys `docs/api` to `gh-pages`. Does that work for you?

I didn't do that at first because I wasn't sure how the deployment for read the docs works and I didn't want to mess with it.

robertmain:
Successfully moved the docs to `api/docs`

davedevelopment:
I think I've set everything up as needed... will soon see.

davedevelopment:
Looks like a problem with the generation itself? https://travis-ci.org/mockery/mockery/jobs/346804425

robertmain:
I honestly have no idea why this is happening. I just cloned a fresh copy of mockery, ran the install and ran the `docs` command. Though it does yell about missing docblocks, it generates a set of documentation in `api/docs`...

davedevelopment:
It's probably just version mismatches, the one that failed is the build where we use the lowest possible versions to satisfy all of the constraints. We'll probably need to bump something somewhere.

davedevelopment:
@robertmain this might do it? https://github.com/mockery/mockery/pull/858

PR #860, opened 2018-03-02
Adds andAnyOther argument matcher

robertbasic:
The name is on purpose "andAnyOther" so that it
is visible from the name itself that it should
never be the first matcher in a list of matchers,
and hopefuly the "other" part says enough that
it will match more than one argument.

### DIFF
--- a/library/Mockery.php
+++ b/library/Mockery.php
@@ -297,13 +297,26 @@ class Mockery
     }
 
     /**
-     * Return instance of AndAnyOther matcher.
+     * Return instance of AndAnyOtherArgs matcher.
      *
-     * @return \Mockery\Matcher\AndAnyOther
+     * An alternative name to `andAnyOtherArgs` so
+     * the API stays closer to `any` as well.
+     *
+     * @return \Mockery\Matcher\AndAnyOtherArgs
      */
-    public static function andAnyOther()
+    public static function andAnyOthers()
     {
-        return new \Mockery\Matcher\AndAnyOther();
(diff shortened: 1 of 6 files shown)


robertbasic:
Related issues: #531 and #301

davedevelopment:
👍think the `and` prefix is a really nice touch. I might be tempted by `andAnyOtherArgs()` though, to go with `anyArgs()`, what do you think? To much of a mouthful?

robertbasic:
@davedevelopment why do we actually have `anyArgs`? What's the difference between that and `any`? It's not documented.

davedevelopment:
@robertbasic I believe `Mockery::any()` will only match one argument at that single position. `anyArgs` will match any call to that method.

It's there because of the new syntax, where we can't differentiate between a call to a method without any arguments and a call where we don't care about the arguments:

``` php

$mock->shouldReceive("foo")->once(); // we don't care about the args, only that it was called.

$mock->expects()->foo(); // we don't have the luxury of not specifying args, so this explicitly expects no args 
$mock->expects()->foo(anyArgs()); // here we don't care about the args

```

robertbasic:
OK, I'll bring it in line then with anyArgs

davedevelopment:
@robertbasic you sure? This is a committee, I'm not BDFL

robertbasic:
@davedevelopment I'm not/I wasn't, so I met you half way :)

Bilge:
👏

PR #869, opened 2018-04-10
Fix iterable return type

Dzov:
Hi,

Since php 7.1 iterable can be used as return type. (http://php.net/manual/en/language.types.iterable.php) 

In 1.0, Mockery tries to create a class name iterable. This fixes this behavior and returns an array instead.

### DIFF
--- a/library/Mockery/Mock.php
+++ b/library/Mockery/Mock.php
@@ -699,7 +699,10 @@ class Mock implements MockInterface
             case 'int':    return 0;
             case 'float':  return 0.0;
             case 'bool':   return false;
-            case 'array':  return [];
+
+            case 'array':
+            case 'iterable':
+                return [];
 
             case 'callable':
             case 'Closure':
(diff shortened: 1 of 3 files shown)


davedevelopment:
Do we need some BC here? Obviously very contrived, but if someone has on their PHP 7.0 app:

``` php
class Thing {
    function getIterable(): iterable {
        return new iterable;
    }
}
```

GrahamCampbell:
After my changes in 1.3, this can now be safely implemented in on the 1.3 branch as a bug fix. This is because in PHP versions where iterable could be the name of an object, it will be resolved to `\iterable` and otherwise to `iterable`, so we can tell the difference. 🎉

GrahamCampbell:
I'll send a PR to do this shortly.

GrahamCampbell:
Done: https://github.com/mockery/mockery/pull/1075.

Issue #871, opened 2018-04-13
Named mock mocking an interface with a return type double-prefixes the namespace to the return type


Bilge:
When creating a named mock that mocks an interface with a return type, Mockery attempts to double-prefix the namespace to the return type, thus always creating an invalid mock and raising a fatal error similar to the following.

>Fatal error: Declaration of ScriptFUSION\Porter\Provider\AsyncProvider5ad0ca3a4a34c::getAsyncConnector(): ScriptFUSION\Porter\Provider\ScriptFUSION\Porter\Connector\AsyncConnector must be compatible with ScriptFUSION\Porter\Provider\AsyncProvider::getAsyncConnector(): ScriptFUSION\Porter\Connector\AsyncConnector in vendor\mockery\mockery\library\Mockery\Loader\EvalLoader.php(34) : eval()'d code on line 28

* Expected type: `ScriptFUSION\Porter\Connector\AsyncConnector`
* Actual type: `ScriptFUSION\Porter\Provider\ScriptFUSION\Porter\Connector\AsyncConnector`
* Only happens with named mocks.
* Only happens on interface methods that define a return type.
* Affects 0.9 and 1.0 branches.

Code used to generate the error:

```php
use ScriptFUSION\Porter\Provider\AsyncProvider;

\Mockery::namedMock(uniqid(AsyncProvider::class, false), AsyncProvider::class)

// AsyncProvider.php
use ScriptFUSION\Porter\Connector\AsyncConnector;

interface AsyncProvider
{
    public function getAsyncConnector(): AsyncConnector;
}
```

PR #872, opened 2018-04-13
Correctly qualifies return types in namespaced mocks

davedevelopment:
Refs #871

### DIFF
--- a/tests/Mockery/NamedMockTest.php
+++ b/tests/Mockery/NamedMockTest.php
@@ -65,4 +65,17 @@ class NamedMockTest extends MockeryTestCase
         );
         $this->assertTrue($gardener->water($cactus));
     }
+
+    /** @test */
+    public function it_gracefully_handles_namespacing()
+    {
+        $animal = Mockery::namedMock(
+            uniqid(Nature\Animall::class, false),
+            Nature\Animal::class
+        );
+
+        $animal->shouldReceive("habitat")->andReturn(new Nature\Habitat());
+
+        $this->assertInstanceOf(Nature\Habitat::class, $animal->habitat());
+    }
 }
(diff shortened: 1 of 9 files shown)


tests/Mockery/NamedMockTest.php
```diff
@@ -65,4 +65,17 @@ public function itCreatesConcreteMethodImplementationWithReturnType()
         );
         $this->assertTrue($gardener->water($cactus));
     }
+
+    /** @test */
+    public function it_gracefully_handles_namespacing()
+    {
+        $animal = Mockery::namedMock(
+            uniqid(Nature\Animall::class, false),
```
Bilge: Typo? `Animall` has two `l`s here.

tests/Mockery/NamedMockTest.php
```diff
@@ -65,4 +65,22 @@ public function itCreatesConcreteMethodImplementationWithReturnType()
...
         $this->assertTrue($gardener->water($cactus));
     }
+
+    /** 
+     * @test 
+     * @requires PHP 7.0.0
+     */
+    public function it_gracefully_handles_namespacing()
+    {
+        require_once __DIR__.'/DummyClasses/NamespacedReturnTypes.php';
```
Bilge: This wouldn't be necessary if you had an `autoload-dev` namespace in `composer.json`, which should be preferred, since having such an definition will preclude the use of this redundant pattern throughout your tests.

davedevelopment:
Argh, sorry about this. I upgraded Mac OS some time ago and all my php binaries need relinking or recompiling, however that works. Will test properly when I get back to my desk.

tests/Mockery/NamedMockTest.php
davedevelopment: 👍

PR #873, opened 2018-04-16
Fixed Travis dependency caching

Bilge:
Actually cache dependencies instead of just pretending, to speed up builds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ matrix:
 
 cache:
   directories:
-    - .composer/cache
+    - vendor
 
 before_install:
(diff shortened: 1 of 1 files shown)


davedevelopment:
I think we probably just need `$HOME/.composer/cache`?

Bilge:
That's not as effective. Composer still has to check out everything from its cache, whereas if it finds everything already checked out in the target `vendor` directory, it does nothing other than regenerate the auto-load file (if everything is still up-to-date). I realise most people cache the Composer cache directory, but I remain certain they're doing it wrong. I implore you to try both if you have any doubts, as I have.

PR #880, opened 2018-05-14
Use TestListenerDefaultImplementation

### LINKED ISSUE #850: Mockery is not compatible with phpunit 7.0
> The `PHPUnit\Framework\BaseTestListener` class has been removed, it was deprecated since PHPUnit 6.4. Please use the `PHPUnit\Framework\TestListenerDefaultImplementation instead`.

barryvdh:
Fixes #850 for PHPUnit 7 compatibility.


barryvdh:
So yeah, this only works for phpunit 7, not 6. I could add a check, but don't know which versions you want to support..

prisis:
phpunit 6 and 7 should work else it is a bc

barryvdh:
Sorry, not really sure why it's failing on that class, when it should use the other one, perhaps because of code coverage.

And probably Travis needs to be update to check all versions, not just oldest and latest.

barryvdh:
This doesn't seem entirely correct. Perhaps it's just easier to create a new major version for phpunit 7, as that adds typehints everywhere, breaking all classes that extend anything.

jmrieger:
@barryvdh - is there currently a roadmap or feature list for 2.0-alpha? I wouldn't mind allocating some effort toward making this PHPUnit7 / PHP7+ compatible

barryvdh:
Ehh, I actually just removed the listener and used the Trait, which is the proper way I guess.

mpjraaij:
How can I install this patch?

PR #886, opened 2018-06-07
Update class_constants.rst

carltondickson:
The first comment here is what I believe the documentation should say

https://github.com/mockery/mockery/pull/829


carltondickson:
From what I can see it's not possible to do this for a class you've overloaded?

davedevelopment:
Thanks! As for the overloaded, I think it should be possible, but I've not tried it or used this feature before.

PR #892, opened 2018-06-07
Fix mocking with classes that use HHVM memoization

karoun:
HHVM has a [built-in decorator](https://docs.hhvm.com/hack/attributes/special#__memoize) that memoizes class methods. However, its implementation is incompatible with the code generation/evaluation strategy that Mockery uses.

This PR fixes Mockery to allow mocking of classes that use the decorator, though with the caveat that the decorator will no longer function as advertised. This is regrettable, but memoization (as implemented by HHVM) is solely a means of improving performance, and so disabling it in tests should not greatly affect behavior.


davedevelopment:
Looks good, please could you squash out those other commits?

karoun:
Replaced by #893 due to my lack of skill with Git.

PR #904, opened 2018-08-17
Add PhpStorm Advanced Metadata for autocompletion

alexdenvir:
I saw a previous attempt at this was made in #759, but it looks like @wan2land hasn't got around to creating a new PR based on the new format for PhpStorm metadata.

This PR is basically the same thing as #759, but using the new format.

For reference, [here is the documentation about this feature in PhpStorm](https://confluence.jetbrains.com/display/PhpStorm/PhpStorm+Advanced+Metadata)

### DIFF
--- a/.phpstorm.meta.php
+++ b/.phpstorm.meta.php
@@ -1,0 +1,11 @@
+<?php
+
+namespace PHPSTORM_META;
+
+override(\Mockery::mock(0), type(0));
+override(\Mockery::spy(0), type(0));
+override(\Mockery::namedMock(0), type(0));
+override(\Mockery::instanceMock(0), type(0));
+override(\mock(0), type(0));
+override(\spy(0), type(0));
+override(\namedMock(0), type(0));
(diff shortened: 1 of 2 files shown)


davedevelopment:
@robertbasic do you use PHPStorm by any chance? This looks fine to me, but another pair of eyes won't hurt.

davedevelopment:
Thank you @alexdenvir  and @AndersonZacharyT :+1:

Bilge:
This doesn't appear to have any effect. Do I have to do something to load it into PhpStorm? Instructions are unclear, but it seems to suggest it should require zero configuration.

alexdenvir:
Still appears to work for me. I would suggest invalidating your IDE cache `File >>> Invalid Caches / Restart >>> Invalidate and Restart`

Bilge:
Invalidating cache has no effect.

jonnyynnoj:
@Bilge Unfortunately it breaks down when used with expectation chaining. If you assign the mock to a variable first `$mock = Mockery::mock(Connector::class);` you will get the metadata.

PR #909, opened 2018-09-27
Upgrading to HHVM 3.25 causes errors due to return type

xxtanisxx:
Return type cannot be properly evaluated by PHP `eval()` function. Stripping out the following return type: `tuple`, `array<*>` or anything with `<*>` in it, `HH\void`, `HH\this`, `HH\ImmVector`, `HH\ImmMap`, `HH\ImmSet`.

The hope is HHVM have a replacement for `eval`. If not, this seems to be the best course forward

- [x] Add smoke test

### DIFF
--- a/.styleci.yml
+++ b/.styleci.yml
@@ -2,5 +2,6 @@ preset: psr2
 
 finder:
   not-name:
+    - MethodWithHHVMReturnType.php
     - MockingParameterAndReturnTypesTest.php
     - SemiReservedWordsAsMethods.php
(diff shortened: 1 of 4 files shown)


davedevelopment:
This looks ok to me for now, please could you add a quick test? Just some sort of smoke test, nothing complicated.

xxtanisxx:
Not too sure what's the best approach for the test. Hope this will suffice. 

The newly added test will pass in hhvm smoke test. Thanks!

davedevelopment:
Looks great to me, thank you 👍

PR #911, opened 2018-09-27
Calls the most common method

davedevelopment:
I think this method is available across the PHPUnits we support...


davedevelopment:
Looks good, the build failure was something in phpdocumentor...

PR #914, opened 2018-10-03
Fix shouldHaveReceived parameters declaration

### LINKED ISSUE #894: PhpStorm: "Required parameter $method missing" error when using alternative syntax
Sample:
```
$spy = Mockery::spy(MyRepo::class);

$serviceA->doThis();

$spy->shouldHaveReceived()->save(.....)
```

I think it's detailed enough. Just say if need more info.

villfa:
According to `MockInterface` the first argument of `shouldHaveReceived()` is required, whereas it can be used with no parameter.
Fix #894.

### DIFF
--- a/library/Mockery/MockInterface.php
+++ b/library/Mockery/MockInterface.php
@@ -99,7 +99,7 @@ interface MockInterface
      * @param null $args
      * @return mixed
      */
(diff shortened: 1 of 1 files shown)


davedevelopment:
Thanks :+1:

Issue #905, opened 2018-08-25
Convenient methods for faking Array interfaces


courtney-miles:
I have a trait I copy across my various projects to make it easy to mock objects that implement the various array interfaces.

It's pretty simple.  You supply it the name of the class to mock, and the array values for it to return for whichever array-like interfaces it may implement.

The only catch is that you can't set expectations.

I use this a lot test code that works with `\Symfony\Component\Validator\ConstraintViolationListInterface`.

Would the maintainers be interested in a PR to include this functionality with Mockery?

This would be inline with the discussion in #63

davedevelopment:
Open the PR and we'll take a look :+1:

At worst, there'll be a working example for others to look at.

PR #916, opened 2018-10-05
Stub array interfaces from an existing array.

courtney-miles:
This PR is for #905 

It provides a convenient way to stub all the methods associated with a class that implements `\Iterator`, `\ArrayAccess` or `\Countable`, by supplying an array as the source for the method values.

```php
$arrOfViolations = [
    \Mockery::mock(\Symfony\Component\Validator\ConstraintViolationInterface::class),
    \Mockery::mock(\Symfony\Component\Validator\ConstraintViolationInterface::class),
];

$mockViolationList = \Mockery::stubTraversable(
    \Symfony\Component\Validator\ConstraintViolationListInterface::class,
    $arrOfViolations
);

echo count($mockViolationList); // 2

foreach ($mockViolationList as $violation) {
    // ...
}
```

A con with this current implementation is that the array items can't be updated.  I.e. it does not stub `ArrayAccess::offsetSet()` and `ArrayAccess::offsetUnset()`

### DIFF
--- a/library/Mockery.php
+++ b/library/Mockery.php
@@ -135,6 +135,17 @@ class Mockery
     }
 
     /**
+     * @param string $class
+     * @param array $items
+     * @return \Mockery\MockInterface
+     * @throws \Mockery\Exception\RuntimeException
+     */
+    public static function stubTraversable($class, array $items)
+    {
+        return self::getContainer()->stubTraversable($class, $items);
+    }
+
+    /**
      * Static and Semantic shortcut to \Mockery\Container::mock().
      *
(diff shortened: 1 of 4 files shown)


courtney-miles:
> Is this a limitation of Mockery or just this specific implementation?

There was an issue with my implementation. I don't recall exactly what the  issue was, but I wanted to see if there was interest in this before I invested in finding a solution.

> I wonder if most of the logic here could be pulled in to it's own class?

I'm happy to shift it. 

Stay tuned for the next update.

PR #917, opened 2018-10-08
Strip HHVM return type and prevent HH object creation

xxtanisxx:
HHVM has type-constant `const type Example = ExampleClass` which allows custom return type  As a result, return type can be anything specified by the developer. Therefore, we strip all return type for HHVM.

Prevent Mockery from creating HH\this and other HH namespace class which already exists for HHVM.

- [x] Need to add test.


davedevelopment:
Thanks! @xxtanisxx are you using HHVM a lot?

PR #927, opened 2018-11-13
[1.0] Unnecessary import removed from the `helpers.php`.

dmitry-ivanov:
That import is not used in the file.


davedevelopment:
Thanks!

PR #929, opened 2018-11-15
Fix in documentation of test doubles

hco:
Maybe I am confused, but the identifier can be a class name, right? 🤔 
I mean, that's how you make it inherit that class - or am I misreading something and identifier is not the thing you pass to `\Mockery::mock`?


davedevelopment:
I'm not all that sure any of that makes sense, but your version is less likely to confuse people, so I'll merge :+1:

PR #930, opened 2018-11-16
Update expectations.rst

rdarcy1:
Makes it clearer that exception name and message should be strings, keeps consistent with `'name_of_method'` usages.


davedevelopment:
Thanks!

PR #931, opened 2018-11-22
Change type of unpacked var

muglug:
It's correct [here](https://github.com/mockery/mockery/blob/dc9255b24b028cb49fae09a8f46345f483114fff/library/Mockery/CompositeExpectation.php#L115) but not in these other two places.

Awesome project!


robertbasic:
Thank you @muglug!

PR #944, opened 2019-02-08
Set property values on all related mocks (0.9 branch)

### LINKED ISSUE #451: Public property is set on wrong instance when using "instance mock"
When trying to set a public property on an instance mock the property is set on the wrong instance.

Expected output:
instance!

Actual output:
mock!

``` php
<?php

require_once __DIR__.'/vendor/autoload.php';

$mock = \Mockery::mock('overload:');
$mock
    ->shouldReceive('foo')
    ->set('bar', true)
;

$instance = new $mock();
$instance->foo();

echo isset($instance->bar) ? "instance!\n" : '';
echo isset($mock->bar) ? "mock!\n" : '';
```

afilina:
When setting values for properties, don't set them just on
the current mock, but set them on all the related mocks that
are in the container.

This way the overloaded instance mocks will get the properties
set, just as the mock itself.

Fixes #451 in the 0.9 branch (for those who can't upgrade to PHP 5.6).

### DIFF
--- a/library/Mockery/Expectation.php
+++ b/library/Mockery/Expectation.php
@@ -192,10 +192,17 @@ class Expectation implements ExpectationInterface
      */
     protected function _setValues()
     {
+        $mockClass = get_class($this->_mock);
+        $container = $this->_mock->mockery_getContainer();
+        $mocks = $container->getMocks();
         foreach ($this->_setQueue as $name => &$values) {
             if (count($values) > 0) {
                 $value = array_shift($values);
-                $this->_mock->{$name} = $value;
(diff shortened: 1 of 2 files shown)


robertbasic:
Thanks Anna!

afilina:
@robertbasic Any chance of a stable release to include this fix?

davedevelopment:
@afilina https://github.com/mockery/mockery/releases/tag/0.9.11

afilina:
@davedevelopment That was fast! Thank you so much!

Issue #946, opened 2019-02-13
PHP 5.6 BC breaking changes in release 1.2.1


MKodde:
Hi! I'm having some difficulties in using the new mockery version in my PHP 5.6 project.

The first thing that seemed to have been introduced in 1.2.1 is the introduction of a strict_type declaration in the [`MockeryPHPUnitIntegrationAssertPostConditionsForV7AndPrevious`](https://github.com/mockery/mockery/blob/master/library/Mockery/Adapter/Phpunit/MockeryPHPUnitIntegrationAssertPostConditionsForV7AndPrevious.php#L21) trait. Introduced in this [commit](https://github.com/mockery/mockery/commit/aa0020fa143c2b06651911c872d300873f8464a4).

In my opinion, the solution would be to remove the strict type declaration, or to update the PHP version restraints, but that would result in the release of a new major version I suppose?

The second is an interesting one, and is related to the [TestListenerTrait](https://github.com/mockery/mockery/blob/master/library/Mockery/Adapter/Phpunit/Legacy/TestListenerTrait.php#L23). My version of PHPUnit (5.7.27) provides some [forwards compatibility features](https://github.com/sebastianbergmann/phpunit/tree/5.7.27/src/ForwardCompatibility). These interfere with the condition mentioned in the TestListenerTrait. For, the second portion of the class alias condition does not resolve to be `true` as the forward compatibility provided by PHPUnit already made the alias for us.

My solution would be to change the second portion of the condition to test for the class existence of `PHPUnit\Util\Blacklist`

I will open a PR shortly with the proposed changes. Please feel free to steer me in a more desired direction.

BackEndTea:
> In my opinion, the solution would be to remove the strict type declaration, or to update the PHP version restraints, but that would result in the release of a new major version I suppose?

Depending on how you interpet semver, bumping the php version does not constitue a bump in major version of a package. Because composer will halt installation of that new version, because it is incompatible with your php version.

It would be bad practice to do this in a patch version, or not communicate this properly though.

davedevelopment:
Fixed in #947, thank you

MKodde:
Thanks for the quick response and reviewing guys, I'm closing this issue!

PR #947, opened 2019-02-13
Provide PHP 5.6 / PHPUnit 5.7.27 support

MKodde:
The current combination of PHPUnit 5.7.27 with Mockery 1.2.1 cause fatal errors which are described in detail in #946. This PR addresses the two issues described in the aforementioned issue.

The commit messages will link to relevant sources and explain the proposed fixes in higher detail.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## x.y.z (unreleased)
 
+* Fix a BC breaking change for PHP 5.6/PHPUnit 5.7.27 (#947) 
+
 ## 1.2.1 (2019-02-07)
 
 * Support for PHPUnit 8 (#942)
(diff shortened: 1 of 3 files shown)


BackEndTea:
A bigger worry here is that travis did not pick up this bc break. Its worth looking into how it is possible that it got through without CI telling us about the BC break

davedevelopment:
Thank you for the patch, will merge.

@BackEndTea In terms of Travis not picking it up, I'm not sure. (https://travis-ci.org/mockery/mockery/jobs/489002164)[This run] (PHP 5.6.32, PHPUnit 5.7.27) warned about the strict types declaration, but didn't fail around the second problem, I'd assume because of the order the autoloader picked things up? 

We're supporting 4 major versions of PHPUnit, and numerous major versions of PHP, things will undoubtedly slip through the net.

PR #953, opened 2019-02-26
Fix type hints in the generator to return an absolute class name

afilina:
When mocking like this: 
```php
$guzzle = Mockery::mock('overload:'.GuzzleClient::class, ClientInterface::class);
```

I get the following error:

```php
Declaration of
GuzzleHttp\Client::send(GuzzleHttp\Psr\Http\Message\RequestInterface $request, array $options = Array)
must be compatible with
GuzzleHttp\ClientInterface::send(Psr\Http\Message\RequestInterface $request, array $options = Array)
```

Notice the first type hint begins with `GuzzleHttp\`. This is because the generator returns relative type hints, which is a bug. This PR fixes it by always return the absolute type hint.

I also added a unit test for this. The rest of the suite passes with this change.

### DIFF
--- a/library/Mockery/Generator/Parameter.php
+++ b/library/Mockery/Generator/Parameter.php
@@ -48,7 +48,7 @@ class Parameter
         if ((version_compare(PHP_VERSION, '5.4.1') >= 0)) {
             try {
                 if ($this->rfp->getClass()) {
-                    return $this->getOptionalSign() . $this->rfp->getClass()->getName();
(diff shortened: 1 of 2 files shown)


afilina:
Looks like this issue might be present on the master branch too. If this PR goes through, I'll write a test and submit the same PR to master. https://github.com/mockery/mockery/blob/master/library/Mockery/Generator/Parameter.php#L69

davedevelopment:
Thanks :+1:

Issue #768, opened 2017-07-24
Multiple demeter chains on multiple objects


mjpvandenberg:
There is a problem when defining multiple demeter chains on multiple mocks. Mockery will not recognize the second expectation of the second mock. Here's a basic unit test.

```php
public function testDemeterChains()
{
    $mock1 = Mockery::mock('object')->shouldIgnoreMissing();
    $mock2 = Mockery::mock('object')->shouldIgnoreMissing();

    $mock1->shouldReceive('getElement->getFirst')->andReturn(1);
    $mock1->shouldReceive('getElement->getSecond')->andReturn(2);
    $mock2->shouldReceive('getElement->getFirst')->andReturn(3);
    $mock2->shouldReceive('getElement->getSecond')->andReturn(4);

    $this->assertEquals(1, $mock1->getElement()->getFirst());
    $this->assertEquals(2, $mock1->getElement()->getSecond());
    $this->assertEquals(3, $mock2->getElement()->getFirst());
    $this->assertEquals(4, $mock2->getElement()->getSecond());
}
```

The last assertion will throw an exception: `Method Mockery_1__demeter_getElement::getSecond() does not exist on this mock object`. I'm using Mockery 0.9.9.

davedevelopment:
Thanks for the detailed report, I have no idea why that might be off the top of my head.

macdabby:
I can confirm I'm getting the same on the current master branch:

        $mock = Mockery::mock('Package\Model1');
        $mock->shouldReceive('select->some->data')->andReturn(1);
        $mock->shouldReceive('select->some->other->data')->andReturn(2);
        $mock = Mockery::mock('overload:Package\Model2');
        $mock->shouldReceive('select->some->data')->andReturn(3);
        $mock->shouldReceive('select->some->other->data')->andReturn(4);

        $out1 = Model1::select()->some()->data();
        $out2 = Model1::select()->some()->other()->data();
        $out3 = Model2::select()->some()->data();
        $out4 = Model2::select()->some()->other()->data();

    Method Mockery_1__demeter_some::other() does not exist on this mock object

macdabby:
I'm not entirely sure how this works, but it almost looks as if the second shouldReceive() call is adding the new chain to the first mock object. It would make sense that some data is bleeding through since these are static methods and likely static vars, but it seems that these are recreated in some eval() process that should isolate them.

robertbasic:
It's not about the order of calls, but the order of setting up expectations.

This breaks:

[...]

Issue #59, opened 2012-04-06
Problem with multiple Demeter chains?


geezmo:
Hello,
i'm having problems with multiple demeter-chain expectations on a mock. 

```
$contextMock = m::mock('\My\Object\Interface);
$contextMock->shouldReceive('getToken->getUser->getCustomers->count')->andReturn(0);
$contextMock->shouldReceive('getToken->getUser->getProfileData->getContactName')->andReturn('foo');
```

The code i'm trying to test is the following:

```
public function doSomething() {
    $user = $myObject->getToken()->getUser()->getCustomers()->count();
    $name = $myObject->getToken()->getUser()->getProfileData()->getContactName();
(...)
}
```

But PHPUnit rises an exception when calling the second function:

```
BadMethodCallException: Method demeter_getUser::getProfileData() does not exist on this mock object
```

If i just invert the two function calls, as well as shouldReceive() calls, PHPUnit claims that the non-existing function is the other one, getCustomers. It seems that only the first chain added is correctly mocked.
Am i missing something? Or is there an issue with multiple demeter chain mocking?
Thanks!

calvinl:
I'm actually still seeing this in the latest version of Mockery (1.1.0). 

With these two demeter chains mocks:

```
$this->braintree->shouldReceive('paymentMethod->find')->once()->andReturn($paymentMethod);
$this->braintree->shouldReceive('paymentMethod->delete')->once()->andReturn($deletePaymentMethod);
```

And I get this error:

```
Method Mockery_4__demeter_3b393f512423a027089d4a2438c220aa_paymentMethod::delete() does not exist on this mock object
```

mjpvandenberg:
This is still/again a problem in v1.2.0.

afeique:
This is still an issue in v1.2.2, PHP 7.0.

I am seeing this:
```
Mockery\Exception\BadMethodCallException : Received demeter_1d88267bbe6a1a5fbf2ab8eb48e7f737_getVolume::getGuid(), but no expectations were specified
```

Which occurs in this test code:
```
$this->image->shouldReceive('getVolume->getGuid')->andReturn($guid)->atLeast()->once();
```

Which is testing this line:
```
$guid = $liveImage->getVolume()->getGuid();
```

With these prototypes:
```
class Image {
    /**
     * @return Volume
     */
    public function getVolume(): Volume
    {
        return $this->volume;
    }
}

class Volume {
    /**
     * @return string
     */
    public function getGuid()
    {
[...]

PR #956, opened 2019-03-07
Fix chained demeter calls with type hint

thiagorb:
The current implementation fails in the following situation:
```
$mock = \Mockery::mock(Order::class);
$mock->shouldReceive('getCustomer->getFirstName')->andReturn('John');
$mock->shouldReceive('getCustomer->getLastName')->andReturn('Doe');
```

In this case, if `getCustomer` has a type hint for return type, the mock for getLastName will not work. This happens because the named mock doesn't have the prefix `Mockery_<count>__`, so instead of `Mockery_1__demeter_02db565163e0992acfde42edfd58d833_getCustomer`, it will be named only `demeter_02db565163e0992acfde42edfd58d833_getCustomer`. 

Because of that, the method `Container::getKeyOfDemeterMockFor` will not be able to find the mock.

Some users complained about this problem in some already closed issues:
- https://github.com/mockery/mockery/issues/59
- https://github.com/mockery/mockery/issues/768

### DIFF
--- a/library/Mockery.php
+++ b/library/Mockery.php
@@ -27,6 +27,7 @@ use Mockery\Loader\EvalLoader;
 use Mockery\Loader\Loader;
 use Mockery\Matcher\MatcherAbstract;
 use Mockery\ClosureWrapper;
+use Mockery\Generator\MockNameBuilder;
 
 class Mockery
 {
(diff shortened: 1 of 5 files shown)


davedevelopment:
Thank you :+1:

PR #967, opened 2019-04-05
Added an argument expectation method to check if expected arguments are parts of real arguments when mocked methods are called

yutod:
This PR is a wrapper of argument expectation.
Already custom validation for arguments can be possible to pass callback function to `withArgs`, which is good!
However, I'd like to suggest the feature to reduce our test code to prepare this wrapper method.
This method helps you to check if expected arguments are actually set when mocked methods are called.

### DIFF
--- a/docs/reference/expectations.rst
+++ b/docs/reference/expectations.rst
@@ -135,6 +135,32 @@ arguments make the closure evaluate to true:
     $mock->foo(4); // matches the expectation
     $mock->foo(3); // throws a NoMatchingExpectationException
 
+Argument matching with some of given values
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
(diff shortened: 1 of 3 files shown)


yutod:
@robertbasic Thanks for reviewing and feedback!
I added some description on docs. I appreciate your feedback again.

robertbasic:
Looking good, thanks @yutod!

yutod:
@davedevelopment 
Thank you for asking @robertbasic to review this PR.
Looks like this changed is confirmed!

If you have further suggestion or any feedback, I'll happily improve it!

Thank you,

yutod:
@davedevelopment 
Thank you for asking @robertbasic to review this PR.
Looks like this changed is confirmed!

If you have further suggestion or any feedback, I'll happily improve it!

Thank you,

Issue #969, opened 2019-04-05
Allow passthru and partial mocking of method handled by __call


tflori:
I know there where issues before for this behavior but I'm wondering why it was never solved. All I read in other issues are workarounds that don't work in all cases.

I found how to solve it but before I open a PR I want to ask what do you think of this solutions.

For partial mocks we would just need to check if there is a call method in the parent class:
```php
        if (!is_null($this->_mockery_partial) && method_exists($this->_mockery_partial, $method) || method_exists($this->_mockery_partial, '__call')) {
            return call_user_func_array(array($this->_mockery_partial, $method), $args);
        } elseif (...) {
          ...
        } elseif ($this->_mockery_deferMissing && method_exists(get_parent_class($this), '__call')) {
            return parent::__call($method, $args);
        } elseif ($this->_mockery_ignoreMissing) {
```

For passthru we would need to check if the method does not exist but a call method exists:
```php
    public function mockery_callSubjectMethod($name, array $args)
    {
        if (!method_exists($this, $name) && method_exists(get_parent_class($this), '__call')) {
            return parent::__call($name, $args);
        }
        return call_user_func_array('parent::' . $name, $args);
    }
```

davedevelopment:
Might be worth a try, I'm not too sure on how it will impact things.

tflori:
merged and released

PR #976, opened 2019-05-14
allow passing thru __call in all mock types

tflori:
See #969

When we see a `__call` method on the parent class we can use this method when php would do the same. This way we are not limiting the developer to not use `__call` method or not mocking classes that use `__call`.

I added tests for all 3 cases. What I didn't touch is the test for the configuration. Mockery might still fail when the mocked class has a `__call` method and the configuration `allowMockingNonExistentMethods` is set to `false`. Typically for php it would return `true` when you ask `is_callable([$object, $method])` when the class has a `__call` method. Just tell me if you want this to be fixed too.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Change Log
 
+* Allow passing thru __call method in all mock types (experimental) (#969)
+
 ## 1.2.2 (2019-02-13)
 
 * Fix a BC breaking change for PHP 5.6/PHPUnit 5.7.27 (#947) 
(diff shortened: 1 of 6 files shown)


davedevelopment:
Sorry for the delay @tflori, I've had some time off work. 

Could you add a note to the `CHANGELOG` describing this and mark it as `EXPERIMENTAL` or something like that?

tflori:
@davedevelopment sure, I will add it on the weekend

GrahamCampbell:
Ping @davedevelopment, @robertbasic. Is there an update here please? Anything I can do to help?

PR #980, opened 2019-06-14
fix compatibility with PHP 7.4

smoench:


smoench:
There is still one error with PHP 7.4 which happens several times:
```
1) Mockery_AdhocTest::testMockeryInterfaceForInterface
Cannot use "parent" when current class scope has no parent
/home/travis/build/mockery/mockery/library/Mockery/Loader/EvalLoader.php:34
/home/travis/build/mockery/mockery/library/Mockery/Loader/EvalLoader.php:34
/home/travis/build/mockery/mockery/library/Mockery/Container.php:230
/home/travis/build/mockery/mockery/tests/Mockery/AdhocTest.php:59
```
But I don't where and how to start debugging. Anyone has a hint?

library/Mockery.php
```diff
@@ -858,7 +858,8 @@ private static function getNewDemeterMock(
                 if ($parRefMethodRetType !== null) {
                     $nameBuilder = new MockNameBuilder();
                     $nameBuilder->addPart('\\' . $newMockName);
-                    $mock = self::namedMock($nameBuilder->build(), (string) $parRefMethodRetType);
+                    $type = PHP_VERSION_ID >= 70100 ? $parRefMethodRetType->getName() : (string)$parRefMethodRetType;
```
GrahamCampbell: Maybe it's worth having a utility method in this library which does this, instead of copying the logic 4 times?

smoench: I think it's not worth it and would be easier to remove them anytime soon when the minimum PHP version gets increased.

smoench:
I could fix the code preventing the `Cannot use "parent" when current class scope has no parent` php deprecation.

GrahamCampbell:
Just tested this with PHP 7.4.0alpha2. Works perfectly. :)

franzliedke:
Good to see this being covered. I also [ran into](https://travis-ci.org/flarum/core/jobs/554884397) the "Cannot use parent..." error.

Thanks for Mockery! :clap:

GrahamCampbell:
Friendly ping @davedevelopment. :)

davedevelopment:
Thank you! Sorry for the delay, my head's a shed at the minute.

PR #985, opened 2019-07-12
Add section for mocking class within class to cookbook

KeyWeeUsr:
Explains mocking in classes with autoloading present by wrapping `new` keyword into function that can be mocked and therefore can return a prepared mock.

### DIFF
--- a/docs/cookbook/index.rst
+++ b/docs/cookbook/index.rst
@@ -11,5 +11,6 @@ Cookbook
     class_constants
     big_parent_class
     mockery_on
+    mocking_class_within_class
 
 .. include:: map.rst.inc
--- a/docs/cookbook/mocking_class_within_class.rst
+++ b/docs/cookbook/mocking_class_within_class.rst
@@ -1,0 +1,146 @@
+.. index::
+    single: Cookbook; Mocking class within class
+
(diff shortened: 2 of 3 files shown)


davedevelopment:
Thank you for this, but I can't help but feel that it probably doesn't belong in the mockery docs.

Will leave it open to garner further opinions.

KeyWeeUsr:
While I agree it's not cleanly mockery-only documentation, it's quite helpful e.g. to people who come from e.g. Python environments or other languages where you can basically mock everything perhaps except a variable in a function/method (local scope). I'm open to reworking the docs because this one example has been bugging me personally for a long while and perhaps the example with GUI elements is not the best example within the PHP world.

robertbasic:
I'm a bit on the fence about this one, but I'd say lets merge it in, @KeyWeeUsr did spend a good time on writing this, plus the cookbook has at least 2 other entries from me that's also in this "grey" area of should it belong in the docs or not.

@davedevelopment what do you think?

davedevelopment:
@robertbasic fire away then :)

robertbasic:
Thanks a bunch @KeyWeeUsr!

PR #988, opened 2019-07-17
Bugfix: __isset call

### LINKED ISSUE #987: call_user_func should be passing $name directly and not wrapping it with an array
https://github.com/mockery/mockery/blob/44a439cc5b0b0bd8a94c9a67dc8801f6cfc9b5b0/library/Mockery/Mock.php#L609

smoench:
resolves #987


davedevelopment:
Thanks!

PR #996, opened 2019-08-27
Bugfix: 'Add support for ! to blacklist methods' for empty methods

lintaba:
it broke our tests:

```
ErrorException: Uninitialized string offset: 0

/var/www/html/src/vendor/sentry/sentry/lib/Raven/Breadcrumbs/ErrorHandler.php:34
/var/www/html/src/vendor/mockery/mockery/library/Mockery/Container.php:155
/var/www/html/src/vendor/mockery/mockery/library/Mockery.php:118
...
```

### DIFF
--- a/library/Mockery/Container.php
+++ b/library/Mockery/Container.php
@@ -149,10 +149,10 @@ class Container
                 }
                 $class = $parts[0];
                 $parts[1] = str_replace(' ', '', $parts[1]);
-                $partialMethods = explode(',', strtolower(rtrim($parts[1], ']')));
+                $partialMethods = array_filter(explode(',', strtolower(rtrim($parts[1], ']'))));
                 $builder->addTarget($class);
                 foreach ($partialMethods as $partialMethod) {
-                    if (!empty($partialMethod) && $partialMethod[0] === '!') {
+                    if ($partialMethod[0] === '!') {
                         $builder->addBlackListedMethod(substr($partialMethod, 1));
                         continue;
                     }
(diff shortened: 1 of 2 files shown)


mfn:
Oh noez, sorry for the break (I added the blacklist stuff).

I wonder though: would purpose has an **empty** `$partialMethod` in the first place? (wasn't aware of this)

They even get added to the "whitelist" (see line below).

Maybe line `152` should be placed not to return an array with empty elements in the first place?

From
```php
$partialMethods = explode(',', strtolower(rtrim($parts[1], ']')));
```
to
```php
$partialMethods = array_filter(explode(',', strtolower(rtrim($parts[1], ']'))));
```
🤷‍♀️

lintaba:
yeah, even better, since in this case this would be still invoked:
```
$builder->addWhiteListedMethod( "" );
```

mfn:
Yeah, I can't think of a use-case for that .. maybe just an unrealized side-effect of the code so far

lintaba:
the behavioral difference may happen in `MockConfiguration::getMethodsToMock`. With blacklisted+whitelisted, it wont care about blacklisteds. My first fix also ignored blacklisted items (since whitelisted is `[""]`, which is not empty). The second fix however respects that, so if you blacklisted methods manually, and used the `Class[]` format, it will respect the blacklisted rules.

davedevelopment:
Thanks all

lintaba:
@davedevelopment (cc @mfn ), what do i need to do to make it part of a release? it still broke our tests :D

davedevelopment:
@lintaba If you add a quick note for this in `CHANGELOG.md` in a PR, I'll tag next week.

lintaba:
> @lintaba If you add a quick note for this in `CHANGELOG.md` in a PR, I'll tag next week.

https://github.com/mockery/mockery/pull/1010

davedevelopment:
@lintaba thank you https://github.com/mockery/mockery/releases/tag/1.2.4

Issue #991, opened 2019-08-08
ExpectationInterface is missing methods that the Expectation class has.


slepic:
Hi,

I have a problem with static analysis of our tests.
We have recently introduced phpstan into our company project. We have succefully fixed all errors in the production code. We have also fixed almost all errors in our tests as well. The only exception are problems with mockery, where methods claim to return some interface (ExpectationInterface), but what they really return is the Expectation class. We are calling methods like once(), with(), etc. on the expectation which works fine and the tests are ok. But the static analysis of the tests code is yelling at us that there are errors calling undefined methods, because the ExpectationInterface does not have those, but the Expectation class does, but the static analysis tool doesnt know about that.
Would it be possible to add these methods on the interface as well? Or what is the reason for those methods being only part of the class and not the interface?

davedevelopment:
I'm not entirely sure if I'm honest! Have you tried adding those methods and running the tests? If nothing breaks, I'd recommend submitting a PR and we could all chime in.

slepic:
Hi, thanks, I will try it and send a PR then, but unfortunately I wont have time for that during at least the next two weeks...

slepic:
Hey there,

so I have created a pull request:
https://github.com/mockery/mockery/pull/998

I have figured out the reason why it wasn't part of the interface.
Reason is that ExpectationInterface is implemented by 2 classes:
- Expectation
- CompositeExpectation

where only the Expectation has those methods,
but CompositeExpectation is relying on the magic __call() to provide those methods, but that is the reason why they weren't on the interface, since the CompositeException provides them only magically.
In the PR I have added those methods on the interface and defined them on the CompositeExpectation class as always delegating to the magic __call method.

I have added methods of the with* family and times* family. I suspect that there might be others though. But these were enough to resolve my issue, so I didnt go any further just yet.

I am perfectly willing to make it complete, but let me first ask, how you wanna proceed with this?
This change is altering an existing interface. Although it seems fine within the library. If someone decided to implement their own ExpectationInterface implementation, their code will break, unless their implementation has those same methods already.
So this should be considered a BC break and as such should be released as next major version (2.0.0).
So unless you want to ignore this possibility and release it as new patch version (likely 1.2.4) I would consider the current state of PR enough to go to such release.
[...]

PR #998, opened 2019-09-05
added methods of Expectation class to ExpectationInterface

slepic:
Resolves issue #991


davedevelopment:
I'm sorry, I missed this when it came in. I'm still not entirely sure if we should proceed with this.

PR #1000, opened 2019-09-06
Improve exception debugging

KmeCnin:
It is often really difficult to understand and be aware of all available handlers for a given method call expectation.
Especially when the difference comes only from an entity inner property.

Here we display a list of all available handlers for the given method and for each one a diff.
Example:
```
1) Foo::testItems
Mockery\Exception\NoMatchingExpectationException: No matching handler found for Mockery_2_Bar::get(object(Hamcrest\Core\IsEqual)). Either the method was unexpected or its arguments matched no expected argument list for this method.
Here is the list of available expectations.

Mockery_2_Bar::get with arguments:
--- Expected
+++ Actual
@@ @@
 Array (
     0 => Hamcrest\Core\IsEqual Object (
         '_item' => Baz Object (
-            'id' => 'foo'
+            'id' => '223918237618995'
         )
     )
 )

Mockery_2_Foo::get with arguments:
--- Expected
+++ Actual
@@ @@
 Array (
     0 => Hamcrest\Core\IsEqual Object (
         '_item' => Bar Object (
-            'id' => 'bar'
+            'id' => '223918237618995'
         )
     )
-    1 => Hamcrest\Core\IsEqual Object (...)
 )
```

Also when the diff consists only in a difference of object reference that would normally result in an empty diff, we display the object ref instead:
```
There was 1 error:

1) Foo::test
Mockery\Exception\NoMatchingExpectationException: No matching expectation found for Mockery_0_Bar::fetch(object(Baz), 'me', ['projection' => '(id)']). Either the method was unexpected or its arguments matched no expected argument list for this method.

Here is the list of available expectations and their diff with actual input:

Mockery_0_Foo::fetch with arguments
--- Expected
+++ Actual
@@ @@
 Array (
-    0 => 'Baz#ref_000000007372a668000000003e36882c'
+    0 => 'Baz#ref_000000007372a677000000003e36882c'
     1 => Hamcrest\Core\IsEqual Object (...)
     2 => Hamcrest\Core\IsEqual Object (...)
 )
```

### DIFF
--- a/library/Mockery/Exception/NoMatchingExpectationException.php
+++ b/library/Mockery/Exception/NoMatchingExpectationException.php
@@ -20,51 +20,186 @@
 
 namespace Mockery\Exception;
 
+use Hamcrest\Util;
 use Mockery;
+use SebastianBergmann\Comparator\ComparisonFailure;
+use SebastianBergmann\Comparator\Factory;
 
 class NoMatchingExpectationException extends Mockery\Exception
 {
-    protected $method = null;
+    /**
+     * @var string
+     */
+    protected $method;
 
-    protected $actual = array();
+    /**
+     * @var array
+     */
+    protected $actual;
 
-    protected $mockObject = null;
+    /**
+     * @var Mockery\MockInterface
+     */
+    protected $mockObject;
 
-    public function setMock(Mockery\LegacyMockInterface $mock)
+    /**
+     * @param string $methodName
+     * @param array $actualArguments
+     * @param array $expectations
+     */
+    public function __construct(
+        Mockery\MockInterface $mock,
+        $methodName,
+        $actualArguments,
+        $expectations
+    ) {
+        $this->setMock($mock);
+        $this->setMethodName($methodName);
+        $this->setActualArguments($actualArguments);
+
+        $diffs = [];
+        foreach ($expectations as $expectation) {
+            $expectedArguments = $expectation->getExpectedArgs();
+
+            $diff = $this->diff(
+                $this->normalizeForDiff($expectedArguments),
+                $this->normalizeForDiff($actualArguments)
+            );
+            if (null === $diff) {
+                // If we reach this, it means that the exception has not been
+                // raised by a non-strict equality. So the diff is null.
+                // We do the comparison again but this time comparing references
+                // of objects.
+                $diff = $this->diff(
+                    $this->normalizeForStrictDiff($expectedArguments),
+                    $this->normalizeForStrictDiff($actualArguments)
+                );
+            }
+
+            $diffs[] = sprintf(
+                "\n%s::%s with arguments%s",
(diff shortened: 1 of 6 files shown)


davedevelopment:
I'm happy with this, @robertbasic please merge if you feel the same way,

library/Mockery/Exception/NoMatchingExpectationException.php
```diff
@@ -20,17 +20,84 @@
 
 namespace Mockery\Exception;
 
+use Hamcrest\Util;
 use Mockery;
+use SebastianBergmann\Comparator\ComparisonFailure;
```
robertbasic: Could you please add the Comparator package then as required dependency to our composer file? Mockery can be used without phpunit, so I'm afraid in those cases this dependency would be most likely missing.

library/Mockery/Exception/NoMatchingExpectationException.php
```diff
@@ -20,17 +20,84 @@
...
+            . ' no expected argument list for this method.'
+            . PHP_EOL . PHP_EOL
+            . 'Here is the list of available expectations and their diff with actual input:'
+            . PHP_EOL
+            . implode('', $diffs);
+
+        parent::__construct($message, 0, null);
+    }
+
+    public function setMock(Mockery\MockInterface $mock)
```
robertbasic: Is this required to be a public method? Can it be private?

library/Mockery/Exception/NoMatchingExpectationException.php
```diff
@@ -67,4 +134,52 @@ public function getMockName()
...
+        try {
+            $comparator->assertEquals($expectedArguments, $actualArguments);
+        } catch (ComparisonFailure $e) {
+            return $e->getDiff();
+        }
+
+        return null;
+    }
+
+    private function normalizeForDiff($args)
```
robertbasic: Could you add docblocks here as well, please?

library/Mockery/Exception/NoMatchingExpectationException.php
```diff
@@ -67,4 +134,52 @@ public function getMockName()
...
+    {
+        // Wraps items with an IsEqual matcher if it isn't a matcher already
+        // in order to be sure to compare same nature objects.
+        return Util::createMatcherArray($args);
+    }
+
+    private function normalizeForStrictDiff($args)
```
robertbasic: Could you add docblocks here as well, please?

library/Mockery/Exception/NoMatchingExpectationException.php
KmeCnin: You are right. Changed for all public methods in this class.

KmeCnin:
Applied your reviews :ok_hand:

robertbasic:
Thanks @KmeCnin!

Issue #992, opened 2019-08-13
andReturnArgument()


downsider:
Hi

In my experience with mockery, a number of times I have found myself needing to return an argument that was passed to a mocked method. In such cases, I have done something like this:

```
$mock->shouldReceive("myMethod")->andReturnUsing(
    function($arg){
        return $arg;
    });
```

This feels really clunky and boilerplatey, but is something commonly required for mocks of classes that do normalisation or follow the decorator pattern.
 
Seeing as PHPUnit has had this functionality for a long time, I feel we should introduce it into Mockery too. I propose we add the following syntax to enable returning arguments

```
$mock->shouldReceive("myMethod")->andReturnArgument(2);
    // where "2" is the index of the argument to return
```

Do you think this is something that can be implemented?

davedevelopment:
I don't see why not, fire away :+1:

downsider:
Hi @davedevelopment I've submitted a PR for this. Could you have a look at the comment I raised please? I'm interested in your opinion on this point.

robertbasic:
Closing as #1001 was merged.

Thank you @downsider!

PR #1001, opened 2019-09-06
andReturnArgument

downsider:
Added functionality to return an argument passed to a method expectation

### DIFF
--- a/library/Mockery/Expectation.php
+++ b/library/Mockery/Expectation.php
@@ -559,6 +559,28 @@ class Expectation implements ExpectationInterface
     }
 
     /**
+     * Sets up a closure to return the nth argument from the expected method call
+     *
+     * @param int $index
+     * @return self
+     */
+    public function andReturnArgument($index)
+    {
+        if (!is_int($index) || $index < 0) {
+            throw new \InvalidArgumentException("Invalid argument index supplied. Index must be a positive integer.");
+        }
+        $closure = function (...$args) use ($index) {
+            if (isset($args[$index])) {
+                return $args[$index];
+            }
+            throw new Exception("Cannot return an argument value. No argument exists for the index $index");
+        };
+
+        $this->_closureQueue = [$closure];
+        return $this;
+    }
+
+    /**
      * Return a self-returning black hole object.
      *
      * @return self
--- a/library/Mockery/Expectation.php
+++ b/library/Mockery/Expectation.php
@@ -573,7 +573,7 @@ class Expectation implements ExpectationInterface
             if (isset($args[$index])) {
                 return $args[$index];
             }
-            throw new Exception("Cannot return an argument value. No argument exists for the index $index");
+            throw new \OutOfBoundsException("Cannot return an argument value. No argument exists for the index $index");
         };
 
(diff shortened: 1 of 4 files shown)


downsider:
Resolves #992

library/Mockery/Expectation.php
```diff
@@ -558,6 +558,28 @@ public function andReturnUsing(...$args)
...
+    public function andReturnArgument($index)
+    {
+        if (!is_int($index) || $index < 0) {
+            throw new \InvalidArgumentException("Invalid argument index supplied. Index must be a positive integer.");
+        }
+        $closure = function (...$args) use ($index) {
+            if (isset($args[$index])) {
+                return $args[$index];
+            }
+            throw new Exception("Cannot return an argument value. No argument exists for the index $index");
```
downsider: I'm not 100% sure if throwing an exception here is correct. 

What do people think is the right behaviour when the argument index is out-of-bounds for a particular method call?

davedevelopment: I think it's a reasonable starting point. Could throw a `OutOfBoundsException` to be more explicit.

library/Mockery/Expectation.php
```diff
@@ -558,6 +558,28 @@ public function andReturnUsing(...$args)
...
+     * @param int $index
+     * @return self
+     */
+    public function andReturnArgument($index)
+    {
+        if (!is_int($index) || $index < 0) {
+            throw new \InvalidArgumentException("Invalid argument index supplied. Index must be a positive integer.");
+        }
+        $closure = function (...$args) use ($index) {
+            if (isset($args[$index])) {
```
davedevelopment: Probably want `array_key_exists` here, the argument could legitimately be `null`. Test case to cover would be good to

library/Mockery/Expectation.php
```diff
@@ -558,6 +558,28 @@ public function andReturnUsing(...$args)
         return $this;
     }
 
+    /**
+     * Sets up a closure to return the nth argument from the expected method call
+     *
+     * @param int $index
+     * @return self
+     */
+    public function andReturnArgument($index)
```
davedevelopment: I'm not 100% sure what would be best, but in other public methods, we use `withArgs`, rather than `withArguments`. 

So would we better with `andReturnArg` or is `Arg` too ambiguous?

davedevelopment:
Sorry for the delay! Missed the notification on this one.

library/Mockery/Expectation.php
downsider: I think `andReturnArg` is fine; I'm happy to keep things consistent.

library/Mockery/Expectation.php
downsider: Good catch, I'll update both

library/Mockery/Expectation.php
downsider: makes sense

downsider:
No worries, thanks for the review

downsider:
@davedevelopment I've updated this MR according to your feedback

davedevelopment:
Thank you, great stuff

PR #1004, opened 2019-09-10
Apply fixes from StyleCI

nathane:
This pull request applies code style fixes from an analysis carried out by [StyleCI](https://github.styleci.io).

---

For more information, click [here](https://github.styleci.io/analyses/z4AxGn).

PR #1007, opened 2019-09-13
disable braces fixer

nathane:
https://github.com/mockery/mockery/pull/1004#issuecomment-530729754
https://github.com/mockery/mockery/pull/1004#issuecomment-530788622

### DIFF
--- a/.styleci.yml
+++ b/.styleci.yml
@@ -5,3 +5,6 @@ finder:
     - MethodWithHHVMReturnType.php
     - MockingParameterAndReturnTypesTest.php
     - SemiReservedWordsAsMethods.php
+
+disabled:
+  - braces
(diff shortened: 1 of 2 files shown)


tests/Mockery/CallableSpyTest.php
```diff
@@ -29,7 +29,8 @@ class CallableSpyTest extends MockeryTestCase
     /** @test */
     public function it_verifies_the_closure_was_called()
     {
-        $spy = spy(function() {});
```
robertbasic: Looks like it still does this weird splitting of braces?

nathane: I've removed outdated commits.👍🏽

robertbasic:
thanks!

Issue #1005, opened 2019-09-11
`andSet()` re-using properties between mock objects.


maccath:
I'm experiencing an issue with Mockery expectations with `andSet`.

**Mockery version:** 1.2.3
**PHPUnit version:** 8.3.4
**PHP version:** 7.3.9

Here is a simple example of the issue:

```
class MockeryTest extends MockeryTestCase
{
    public function testMockery()
    {
        $mockClass = \Mockery::mock('FakeClass');
        $mockClassTwo = \Mockery::mock('FakeClass');

        $mockClass->shouldReceive('setAttribute')
            ->with('name', 'Katy')
            ->andSet('name', 'Katy');

        $mockClassTwo->shouldReceive('setAttribute')
            ->with('name', 'Mike')
            ->andSet('name', 'Mike');

        $mockClass->setAttribute('name', 'Katy');
        $mockClassTwo->setAttribute('name', 'Mike');

        $this->assertEquals('Katy', $mockClass->name);
        $this->assertEquals('Mike', $mockClassTwo->name);
    }
}
```

The tests fail with:

```
Failed asserting that two strings are equal.
Expected :'Katy'
Actual   :'Mike'
```

It seems as though two mocks of the same class are treated as the same object, and properties set on them via `andSet` are shared.

If I set the properties on the mock objects directly (e.g. with `$mockClass->name = 'Katy'`), the tests pass. Similarly, if I use two different class names (e.g. `'FakeClass'` and `'FakeClassTwo`'), the tests pass. It also works if I use named mocks.

Is this intentional, or a bug? If it is intentional, is the only workaround to use named mocks?

robertbasic:
Fixed with #1012, thank you @maccath!

PR #1012, opened 2019-10-03
#1005 Set properties correctly for different mocks of the same class.

maccath:
When using `andSet()` expectation, set the property only on instances of the current Mock rather than _all_ mocks of the same class.

Fixes issue #1005, which appears to have been inadvertently caused when fixing #451 

## Changes:
- Add `_mockery_instanceMock` property and `mockery_isInstance()` method.
  - Set `_mockery_instanceMock` property on `mockery_init()` using config value.
  - Default value is `true` if not passed (preserves existing behaviour).
  - Public method `mockery_isInstance()` returns value of protected `_mockery_instanceMock`
- Only set values for current mock + instances of it.

### DIFF
--- a/library/Mockery/Container.php
+++ b/library/Mockery/Container.php
@@ -230,7 +230,7 @@ class Container
         $this->getLoader()->load($def);
 
         $mock = $this->_getInstance($def->getClassName(), $constructorArgs);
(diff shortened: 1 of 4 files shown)


davedevelopment:
Looks good to me

PR #1014, opened 2019-10-04
Move constructor expectations info to docs

davedevelopment:
As it's not a practice I would like to promote, I'm moving this to the
docs, rather than have it on the README. I think for the future I would
prefer the README to be the Mockery - The Right Way type thing, the docs
contain all the juicy bits you can cut your hands on.

### DIFF
--- a/README.md
+++ b/README.md
@@ -258,21 +258,6 @@ $double->allows()->doFoo()->andReturns(123);
 $double->foo(); // int(123)
 ```
 
-### Testing the constructor arguments of hard Dependencies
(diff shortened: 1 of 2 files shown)


mfn:
I find this good, it's more consistent.

I have no problem (and personally prefer) it e.g. the README would contain the _complete_ documentation (out of question here, just mentioning this).

But it should be clear where is what, or rather: everything relevant should be in one place!

robertbasic:
Good one!

PR #1015, opened 2019-10-14
docs: fix a typo in class constants section

ivastly:
I found on obvious typo in the docs ('mock' call was used instead of a 'namedMock'), so here is the fix! Let me know if you have any comments, thank you!


davedevelopment:
Thank you!

PR #1021, opened 2019-11-07
allow code completion for andReturnSelf()

kohlerdominik:
This PR allows code completion for the `andReturnSelf()`-method by adding the `MockInterface` as return type.
It uses the same return types as the `getMock()`-method.

---

For this example, code completion would now the way we'd expect:

```php
$builder->shouldReceive('addThis')
	->andReturnSelf()
	->shouldReceive('addThat')
	->with(13)
	->andReturnSelf()
	->shouldReceive('get')
	->with(23)
	->andReturn($object);
```
_updated to add parameter example_

### DIFF
--- a/library/Mockery/Expectation.php
+++ b/library/Mockery/Expectation.php
@@ -527,7 +527,7 @@ class Expectation implements ExpectationInterface
     /**
      * Return this mock, like a fluent interface
      *
-     * @return self
+     * @return self|LegacyMockInterface|MockInterface
      */
     public function andReturnSelf()
     {


robertbasic:
Wouldn't you do this with a [demeter chain mock](http://docs.mockery.io/en/latest/reference/demeter_chains.html)?

kohlerdominik:
> 
> 
> Wouldn't you do this with a [demeter chain mock](http://docs.mockery.io/en/latest/reference/demeter_chains.html)?

Thanks for your input!
I stripped my example a little to much. My use case would be with parameters, which isn't supported by the demeter chain:

```
$builder->shouldReceive('addThis')
	->andReturnSelf()
	->shouldReceive('addThat')
	->with(13)
	->andReturnSelf()
	->shouldReceive('get')
	->with(23)
	->andReturn($object);
```

davedevelopment:
Not sure on this one. It works in practice, but only because you usually receive a `CompositeExpectation` from `Mock::shouldReceive`. `CompositeExpectation` has the `__call` to pass unknown calls back to the mock.

PR #1030, opened 2019-12-04
Remove comments from mock pass

robertbasic:
While debugging the memory leak issues, I've noticed that Mockery is doing a bit more work than really necessary. The Mock class is full of docblocks and comments that increase the generated code, meaning the string manipulation passes need more memory.

I'm not really sure about this approach. I've also learned that php has `-w` switch that strips away all comments and whitespace and creates output that can be used instead of Mock.php. The problem with that approach though, that we should be generating that file ourselves for every release (probably can be automated, but still, it's a step that can go wrong.) This StrippedMock.php has better performance impacts than the changes from this MR.

Thoughts?

### DIFF
--- a/library/Mockery/Generator/StringManipulation/Pass/ClassNamePass.php
+++ b/library/Mockery/Generator/StringManipulation/Pass/ClassNamePass.php
@@ -32,18 +32,20 @@ class ClassNamePass implements Pass
 
         $className = $config->getShortName();
 
-        $code = str_replace(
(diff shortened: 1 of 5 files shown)


GrahamCampbell:
I doubt there's better performance. Do you have any evidence? I actually suspect this is slower, because you have to do work to remove the comments Also, stripping phpdoc can be dangerous, because phpdoc can have semantic meaning, such as in doctrine.

PR #1031, opened 2019-12-20
Test on PHP 7.4 and drop HHVM

GrahamCampbell:
1. Test on stable PHP 7.4.
2. Allow composer to actually proceed on PHP 8.
3. Neither HHVM 3 or 4 seem to be able to even start running the tests, so I have dropped HHVM from the build.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,8 @@
-sudo: false
-
 language: php
 
 matrix:
   allow_failures:
-    - php: hhvm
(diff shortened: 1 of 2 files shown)


GrahamCampbell:
Alright, this is ready for review now, All green. :)

// cc @davedevelopment, @robertbasic

davedevelopment:
Thank you @GrahamCampbell

PR #1039, opened 2020-02-14
Fix mocking with anonymous classes

villfa:
Mockery throws a PHP Parse error when trying to mock an anonymous class.

This PR fixes this issue.


davedevelopment:
Thank you :+1:

PR #1042, opened 2020-03-12
Fix PHPDOC since mixed[] indicates only an array of mixed values

briedis:
Newer PhpStorm versions are more strict and now are warning that types don't match.
image

Which is basically true, since `[]` indicates that only arrays are expected.

This PR allows single mixed value as a parameter too.

### DIFF
--- a/library/Mockery/Expectation.php
+++ b/library/Mockery/Expectation.php
@@ -406,7 +406,7 @@ class Expectation implements ExpectationInterface
     /**
      * Expected argument setter for the expectation
      *
-     * @param mixed[] ...$args
+     * @param mixed|mixed[] ...$args
      * @return self
      */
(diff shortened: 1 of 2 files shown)


davedevelopment:
Thanks!

cino:
@davedevelopment Could this be tagged in a patch so I could live a peaceful life without phpstorm warnings? :)

GrahamCampbell:
You can just install the dev version in the meanwhile.

GrahamCampbell:
`^1.3.2` along with global allowing of dev stability with prefer stable will do what you want. Until a stable 1.3.2 tag is there, composer will go get the code from the master branch, and afterwards, it will only get stable tags, 1.3.2 or higher.

PR #1043, opened 2020-03-12
Switch to PSR-12

GrahamCampbell:


.styleci.yml
```diff
@@ -1,10 +1,7 @@
...
+preset: psr12
 
 finder:
   not-name:
     - MethodWithHHVMReturnType.php
     - MockingParameterAndReturnTypesTest.php
     - SemiReservedWordsAsMethods.php
-
-disabled:
-  - braces
```
GrahamCampbell: PSR-12 now allows single line style Closures, so we don't need to disable this any more.

davedevelopment: :tada: woohoo

PR #1044, opened 2020-04-04
Slim down mockery when loaded as a dependency

axlon:
This pull request slims down Mockery as a dependency by removing all unnecessary files from distributions. This means that if you install Mockery via Composer you will no longer have Mockery's tests and meta files in your vendor directory.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,12 @@
+* text=auto
+
+/docker export-ignore
+/tests export-ignore
+.gitattributes export-ignore
+.gitignore export-ignore
+.php_cs export-ignore
+.scrutinizer.yml export-ignore
+.styleci.yml export-ignore
+.travis.yml export-ignore
+Makefile export-ignore
+phpunit.xml.dist export-ignore


davedevelopment:
We've seen this a couple of times in #610 and #641 and I turned it down, but would be happy to know what the general consensus is...

axlon:
My personal reasoning:
I was writing a test and as I typed `extends Test` a bunch of Mockery's own tests showed up, rather than what I was looking for (which was `MockeryTestCase`). 

From my point of view there is no benefit to having Mockery's tests be included in vendor, because you will never run them from there. Say I would want to make a change to Mockery's code I wouldn't be doing that from my project's vendor directory.

mfn:
I agree with the PR FWIW; I support to slim down packages to their relevant runtime parts only (and some very useful meta files like readme/changelog/etc/.)

martinssipenko:
I think this makes change sense.

simonschaufi:
i just wanted to create the same PR for the same reason :smile:

simonschaufi:
@davedevelopment @mfn ping

davedevelopment:
👍 I still don't like it, but I think it makes sense to move forward this way.

Issue #1048, opened 2020-04-22
Uncaught Error: Access to undeclared static property: PHPUnit\Util\Blacklist::$blacklistedClassNames


DemonTPx:
I've upgraded phpunit to version 9.1.2 and get this error:

    PHP Fatal error:  Uncaught Error: Access to undeclared static property: PHPUnit\Util\Blacklist::$blacklistedClassNames in /home/bert/workspace/util-bundle/vendor/mockery/mockery/library/Mockery/Adapter/Phpunit/Legacy/TestListenerTrait.php:88

It seems that it has been moved to `PHPUnit\Util\Blacklist::BLACKLISTED_CLASS_NAMES` and in now a `const` instead of a variable.

See: https://github.com/sebastianbergmann/phpunit/commit/2b4d7b8bb3fd052095e64999990cc36b63aa0771

I searched through the mockery code and it tries to add the `\Mockery` class to this blacklist, which is not possible when its a `const` now...

Downgrading to phpunit 9.1.1 or earlier fixes the error.

PR #1049, opened 2020-04-22
WIP PHPUnit 9 support

davedevelopment:
Refs #1048

I think it's about time to stop supporting older versions of PHPUnit now too, the list is getting long...


davedevelopment:
Superceded by #1053

PR #1053, opened 2020-04-29
Add support for PHPUnit version 9

### LINKED ISSUE #1048: Uncaught Error: Access to undeclared static property: PHPUnit\Util\Blacklist::$blacklistedClassNames

martinssipenko:
This adds support for PHPUnit version 9.

PHPUnit version 9.1.2 had a breaking change in then `@internal` marked `Blacklist` class, which was addressed in [9.1.3](https://github.com/sebastianbergmann/phpunit/blob/74a7e9b91281f5051da617fdb2201c0d619b9efe/ChangeLog-9.1.md#913---2020-04-23).

Fixes #1048

### DIFF
--- a/library/Mockery/Adapter/Phpunit/MockeryTestCase.php
+++ b/library/Mockery/Adapter/Phpunit/MockeryTestCase.php
@@ -38,7 +38,7 @@ abstract class MockeryTestCase extends \PHPUnit\Framework\TestCase
     {
     }
 
-    public function expectExceptionMessageRegExp($regularExpression)
+    public function expectExceptionMessageRegEx($regularExpression)
     {
         if (method_exists(get_parent_class(), 'expectExceptionMessageRegExp')) {
             return parent::expectExceptionMessageRegExp($regularExpression);
--- a/library/Mockery/Adapter/Phpunit/MockeryTestCase.php
+++ b/library/Mockery/Adapter/Phpunit/MockeryTestCase.php
@@ -47,7 +47,7 @@ abstract class MockeryTestCase extends \PHPUnit\Framework\TestCase
         return $this->expectExceptionMessageMatches($regularExpression);
     }
 
-    public static function assertMatchesRegularExpression(string $pattern, string $string, string $message = ''): void
+    public static function assertMatchesRegEx($pattern, $string, $message = '')
     {
         if (method_exists(get_parent_class(), 'assertMatchesRegularExpression')) {
             parent::assertMatchesRegularExpression($pattern, $string, $message);
--- a/library/Mockery/Adapter/Phpunit/MockeryTestCase.php
+++ b/library/Mockery/Adapter/Phpunit/MockeryTestCase.php
@@ -46,4 +46,13 @@ abstract class MockeryTestCase extends \PHPUnit\Framework\TestCase
 
         return $this->expectExceptionMessageMatches($regularExpression);
     }
+
+    public static function assertMatchesRegularExpression(string $pattern, string $string, string $message = ''): void
+    {
+        if (method_exists(get_parent_class(), 'assertMatchesRegularExpression')) {
+            parent::assertMatchesRegularExpression($pattern, $string, $message);
+        }
+
+        self::assertRegExp($pattern, $string, $message);
+    }
 }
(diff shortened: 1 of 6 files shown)


tests/Mockery/Adapter/Phpunit/TestListenerTest.php
```diff
@@ -96,8 +96,24 @@ public function testMockeryIsAddedToBlacklist()
     {
         $suite = \Mockery::mock(\PHPUnit\Framework\TestSuite::class);
 
-        $this->assertArrayNotHasKey(\Mockery::class, \PHPUnit\Util\Blacklist::$blacklistedClassNames);
-        $this->listener->startTestSuite($suite);
-        $this->assertSame(1, \PHPUnit\Util\Blacklist::$blacklistedClassNames[\Mockery::class]);
+        if (method_exists(\PHPUnit\Util\Blacklist::class, 'addDirectory')) {
```
martinssipenko, line 99: Perhaps it would be better to split this up into two tests and skip one of them depending on PHPUnit version?

library/Mockery/Adapter/Phpunit/MockeryTestCase.php
```diff
@@ -37,4 +37,22 @@ protected function mockeryTestSetUp()
     protected function mockeryTestTearDown()
     {
     }
+
+    public function expectExceptionMessageRegEx($regularExpression)
```
davedevelopment, line 41: Would it have been better to leave the method name as `expectExceptionMessageRegExp` here? Or was there a clash in the signatures over the course of different PHPUnit versions? 

The travis build is failing because there are still some tests in there call `expectExceptionMessageRegExp`.

davedevelopment: That's what I did here https://github.com/mockery/mockery/pull/1049/files#diff-491577b1fec453820a42849002bd0828R41

tests/Mockery/Adapter/Phpunit/TestListenerTest.php
davedevelopment: I would agree :+1:

library/Mockery/Adapter/Phpunit/MockeryTestCase.php
martinssipenko: @davedevelopment Yes there are issues with signatures. I'll try to fix travis build so it passes.

tests/Mockery/Adapter/Phpunit/TestListenerTest.php
martinssipenko: @davedevelopment I can't find an easy way to check the minimum required version of PHPUnit is `9.1.3`, so maybe we can leave this as is?

martinssipenko:
@davedevelopment the build is now green, could you please take a look at this again? I would love to see this released as it currently prevents people from upgrading to PHPUnit version above 9.1.2.

davedevelopment:
@martinssipenko I was just about to pull this branch and finish it myself! Will review now.

davedevelopment:
Merged. I'm going to take a look at dropping our list of supported PHP and PHPUnit versions today and will probably make a release later.

martinssipenko:
Sweet, yeah the list of supported PHP version is quite long and is starting to cause much pain.

davedevelopment:
PR created #1059

One thing I'm going to revert from this, those helpers for the regExp methods should be on a test case specifically for the mockery tests, rather than the `MockeryTestCase`, which is for consumers first and foremost. We don't want to have to support those methods ourselves.

martinssipenko:
Maybe they are not needed at all if support for older version, where they were not present or were named differently, is dropped.

PR #1054, opened 2020-05-04
Fix DocBlocks for variadic arguments

thiagorb:
The PHPDoc syntax for variadic arguments is not documented, but seems like most tools settled with this syntax:
```
/**
 * @param string ...$args
 */
function variadic(string ...$args)
{
}
```


davedevelopment:
Thank you :+1:

Issue #1057, opened 2020-05-12
Incorrect Type Hint for shouldNotReceive()


danielsd:
The argument type hint for [shouldNotReceive()](https://github.com/mockery/mockery/blob/master/library/Mockery/Mock.php#L282) is currently `array` which is not entirely correct since the actual argument is `...$methodNames` which takes one or many arguments. This triggers a warning inspection in PhpStorm and possibly other IDEs.

This type hint should probably be changed to `mixed` like the method [shouldReceive()](https://github.com/mockery/mockery/blob/master/library/Mockery/Mock.php#L194).

Proposed PR: https://github.com/mockery/mockery/pull/1058

danielsd:
Fixed by #1058

PR #1058, opened 2020-05-12
Fix type hint on shouldNotRecieve() argument

danielsd:
Addresses https://github.com/mockery/mockery/issues/1057

### DIFF
--- a/library/Mockery/Mock.php
+++ b/library/Mockery/Mock.php
@@ -279,7 +279,7 @@ class Mock implements MockInterface
     /**
      * Shortcut method for setting an expectation that a method should not be called.
      *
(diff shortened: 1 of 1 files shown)


library/Mockery/Mock.php
```diff
@@ -279,7 +279,7 @@ public function expects($something = null)
     /**
      * Shortcut method for setting an expectation that a method should not be called.
      *
-     * @param array ...$methodNames one or many methods that are expected not to be called in this mock
+     * @param mixed ...$methodNames one or many methods that are expected not to be called in this mock
```
davedevelopment, line 282: I think should probably be changed to `string ...$methodNames` and the same done to `shouldReceive`?

davedevelopment:
Thank you!

PR #1059, opened 2020-05-14
Drop support for PHP < 7.3 and PHPUnit < 8

davedevelopment:
The list is getting long and unwieldy.


composer.json
```diff
@@ -31,12 +31,12 @@
...
     ],
     "require": {
-        "php": ">=5.6.0",
+        "php": ">=7.3.0",
         "lib-pcre": ">=7.0",
         "hamcrest/hamcrest-php": "~2.0"
     },
     "require-dev": {
-        "phpunit/phpunit": "~5.7.10|~6.5|~7.0|~8.0|~9.0"
+        "phpunit/phpunit": "~8.0|~9.0"
```
localheinz, line 39:
```suggestion
        "phpunit/phpunit": "^8.0.0 || ^9.0.0"
```

davedevelopment:
@GrahamCampbell [noted](https://twitter.com/GrahamJCampbell/status/1260879743332814848) that this technically requires a major release, as we don't `require` PHPUnit.

I don't have the time or inclination to go and do all the `Towards 2.0` type things now, so I might release 2.0 as 1.0 without some of the garbage.

composer.json
```diff
@@ -31,12 +31,12 @@
         }
     ],
     "require": {
-        "php": ">=5.6.0",
+        "php": ">=7.3.0",
```
localheinz, line 34:
```suggestion
        "php": "^7.3.0",
```

davedevelopment:
Thank you @localheinz

library/Mockery.php
```diff
@@ -88,12 +88,9 @@ public static function builtInTypes()
             'string',
             'iterable',
             'void',
+            'object',
```
localheinz, line 91: What do you think about keeping this list sorted by name?

We can probably also remove the comment `// Up to php 7` that is slightly out of view.

library/Mockery/Adapter/Phpunit/TestListener.php
```diff
@@ -20,16 +20,36 @@
...
+    private $trait;
+
+    public function __construct()
+    {
+        $this->trait = new TestListenerTrait();
+    }
+
+
+    /**
+     * {@inheritdoc}
```
localheinz, line 41: Do we need these DocBlocks?

I believe they are inherited automatically.

library/Mockery/Adapter/Phpunit/TestListener.php
```diff
@@ -20,16 +20,36 @@
...
+    use TestListenerDefaultImplementation;
+
+    private $trait;
+
+    public function __construct()
+    {
+        $this->trait = new TestListenerTrait();
+    }
+
+
```
localheinz, line 39: Looks like we have an extra empty line here, let me open up a pull request enabling a corresponding fixer.

library/Mockery/Adapter/Phpunit/TestListener.php
davedevelopment: Done

library/Mockery/Adapter/Phpunit/TestListener.php
mfn: Yeah, `no_extra_consecutive_blank_lines` isn't part of PSR2 :-/

PR #1062, opened 2020-05-14
Enhancement: Normalize composer.json

localheinz:
This PR

* [x] requires `ergebnis/composer-normalize`
* [x] runs `composer normalize --dry-run` on Travis CI`
* [x] runs `composer normalize`

💁‍♂️ For reference, see https://github.com/ergebnis/composer-normalize.


GrahamCampbell:
:-1: I don't think this is needed. Just gonna cause merge conflicts for no reason.

mfn:
Also: unless some CI job gets created, pretty sure this gets "un-normalized" with the next change

localheinz:
@mfn 

Can be added, but before adding a job, I'd prefer to see if there is interest at all.

🤓

Issue #1056, opened 2020-05-09
Proposal: Verify that quick definitions for default expectations are called at least once


danydev:
I spent quite a bit of time using Mockery in several projects in the last few years, and one of the pain points shared across those is that developers tend to use quick definitions for expectations for cases where that expectation is not crucial for the test. Then time passes, code refactoring come, tests pass but maybe some of those expectations become useless. This means useless expectations accumulate over time in our unit tests.

I think there are still a quite number of cases where quick definitions are a good practice, using them in the right way makes tests clearer with less cluttering for not-so-relevant expectations, making easier to understand the test itself.

For example a developer may use it to mock the 'log' method of a logger that is called in a test, but for which the developer didn't really care to test this behaviour.
```
function test_status_is_new_after_creation() {
    $logger = \Mockery::mock(Logger::class, ['log' => null]);
    $SUT = new SUT($logger);
    $status = $SUT->getStatus($event);
    $this->assertSame(Status::NEW, $status);
}
``` 

Or maybe the developer is using it to 'connect' some mocks together, for which would be impossible for the test to pass if such connection is not in place:
```
function test_it_retrieves_users() {
    $user = \Mockery::mock(User::class, ['getId' => 1234]); // Assuming getId is used internally in SUT to index users by id
    $SUT = new SUT([$user]);
    $brokenUsers = $SUT->getUsers($user);
    $this->assertSame(1234, $brokenUsers[0]);
}
``` 

 Or maybe a developer can use it for some methods for which the response is not relevant to the specific test he's doing, but still needs to be called and thus mocked. 
```
// Imagine a Product class with this method
function canSell(): bool
{
    if ($this->inventoryManager->isSoldOut()) {
        return false;
    }
    if ($this->date > $this->availabilityDate()) {
        return false;
    }
    return true;
}

function test_cannot_sell_when_date_greater_than_availability_date() {
    $inventoryManager = \Mockery::mock(InventoryManager::class, ['isSoldOut' => true]); // <-- yes, right now we need this method to return true to reach the branch we want to test, but maybe in the future we'll move the inventory check below, and this mock will not be relevant for the test
    $SUT = new SUT($inventoryManager, new DateTime('tomorrow'), new DateTime('today'));
    $canSell = $SUT->canSell();
    $this->assertSame(false, $canSell);
}
``` 

Sometimes you don't need to verify that an expectation returns a specific value, the test still makes sense because something else more relevant to what you want to test happened. 
```
[...]

PR #1063, opened 2020-05-21
Add option to allow\disallow mocking methods unnecessarily for default expectations defined through quick definitions

danydev:
…t expectations set through quick definitions

### DIFF
--- a/library/Mockery/Configuration.php
+++ b/library/Mockery/Configuration.php
@@ -41,6 +41,11 @@ class Configuration
     protected $_allowMockingMethodsUnnecessarily = true;
 
     /**
+     * @var QuickDefinitionsConfiguration
+     */
+    protected $_quickDefinitionsConfiguration;
+
+    /**
      * Parameter map for use with PHP internal classes.
      *
      * @var array
@@ -56,6 +61,11 @@ class Configuration
      * @see https://github.com/mockery/mockery/issues/268
      */
     protected $_reflectionCacheEnabled = true;
+
+    public function __construct()
+    {
+        $this->_quickDefinitionsConfiguration = new QuickDefinitionsConfiguration();
+    }
 
     /**
      * Set boolean to allow/prevent mocking of non-existent methods
@@ -152,6 +162,14 @@ class Configuration
     public function getConstantsMap()
     {
         return $this->_constantsMap;
+    }
+
+    /**
+     * Returns the quick definitions configuration
+     */
+    public function getQuickDefinitions(): QuickDefinitionsConfiguration
+    {
+        return $this->_quickDefinitionsConfiguration;
     }
 
     /**
--- a/library/Mockery/Container.php
+++ b/library/Mockery/Container.php
@@ -233,7 +233,11 @@ class Container
         $mock->mockery_init($this, $config->getTargetObject(), $config->isInstanceMock());
 
         if (!empty($quickdefs)) {
-            $mock->shouldReceive($quickdefs)->byDefault();
+            if (\Mockery::getConfiguration()->getQuickDefinitions()->shouldBeCalledAtLeastOnce()) {
+                $mock->shouldReceive($quickdefs)->atLeast()->once();
+            } else {
+                $mock->shouldReceive($quickdefs)->byDefault();
+            }
         }
         if (!empty($expectationClosure)) {
             $expectationClosure($mock);
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Change Log
 
+## 1.4.1 (XXXX-XX-XX)
+
+* Allow quick definitions to use 'at least once' expectation
+  `\Mockery::getConfiguration()->getQuickDefinitions()->shouldBeCalledAtLeastOnce(true)` (#1056)
+
 ## 1.4.0 (2020-05-19)
 
 * Fix andAnyOthers() to properly match earlier expectations (#1051)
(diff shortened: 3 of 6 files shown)


library/Mockery/Container.php
```diff
@@ -233,7 +233,11 @@ public function mock(...$args)
         $mock->mockery_init($this, $config->getTargetObject(), $config->isInstanceMock());
 
         if (!empty($quickdefs)) {
-            $mock->shouldReceive($quickdefs)->byDefault();
+            if (\Mockery::getConfiguration()->getQuickDefinitionsApplicationMode() === \Mockery\Configuration::QUICK_DEFINITIONS_APPLICATION_MODE_DEFAULT_EXPECTATION) {
```
robertbasic, line 236: Rather turn these if statements into questions on the configuration, like:

```
if (\Mockery::getConfiguration()->theQuestionMethodGoesHere()) {
$mock->...->byDefault();
} else {
$mock->...->atLeast()->once();
}
```

now as for the `theQuestionMethodGoesHere`... we should come up with a descriptive method name. `shouldQuickDefinitionSetupAStub()`, `shouldQuickDefinitionSetupAMock()`, `isQuickDefintionModeDefault()`, `shouldSetExpectationOnQuickDefinition()`... Something like this.

robertbasic: Thing is the class constants are really long and make the code a bit unreadable, so we should come up with a better way of setting either of the two modes on the configuration, too.

library/Mockery/Configuration.php
```diff
@@ -103,6 +113,33 @@ public function mockingMethodsUnnecessarilyAllowed()
...
+     *
+     * @param string $applicationMode
+     */
+    public function setQuickDefinitionsApplicationMode($applicationMode)
+    {
+        if (!in_array($applicationMode, array(
+            self::QUICK_DEFINITIONS_APPLICATION_MODE_DEFAULT_EXPECTATION,
+            self::QUICK_DEFINITIONS_APPLICATION_MODE_MOCK_AT_LEAST_ONCE
+        ))) {
+            throw new \InvalidArgumentException('Unexpected application mode: ', $applicationMode);
```
robertbasic, line 128: Test for this case please as well.

library/Mockery/Configuration.php
```diff
@@ -40,6 +43,13 @@ class Configuration
      */
     protected $_allowMockingMethodsUnnecessarily = true;
 
+    /**
+     * Defines what a quick definition should produce.
```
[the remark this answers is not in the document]
robertbasic: And we should document the feature as well :)

library/Mockery/Configuration.php
```diff
@@ -22,6 +22,9 @@
 
 class Configuration
 {
+    const QUICK_DEFINITIONS_APPLICATION_MODE_MOCK_AT_LEAST_ONCE = 'QUICK_DEFINITIONS_APPLICATION_MODE_MOCK_AT_LEAST_ONCE';
```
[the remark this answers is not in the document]
robertbasic: Also, do we need the word "application" in there? Feels like it's just extra "noise".

library/Mockery/Container.php
danydev: I see your point, before doing the changes, I would like to offer my point of view.
When I chose to use a 'setting' configuration over a boolean I did so, because I thought that a boolean would not model perfectly this behaviour, and it's not really extendable if we decide to add other behaviour in the future.

Indeed if we use both `shouldQuickDefinitionSetupAStub` and `shouldQuickDefinitionSetupAMock` we are not going to express that those are mutually exclusive from an API point of view. If we use just one (let' say just `shouldQuickDefinitionSetupAStub`), we're not descriptive enough about what (and why) should happen when it's false.

Still I understand how the longness and naming is not cool. What I would usually do in my code in those cases is to use an [enum-like](https://github.com/myclabs/php-enum) class that would look something like `QuickDefinitionMode` having as "values" `DEFAULT` and `MOCK_AT_LEAST_ONCE` at this point the getter and setter would look something like \Mockery\Configuration::setQuickDefinitionMode(QuickDefinitionMode::MOCK_AT_LEAST_ONCE())`. I think the lenght is fair enough and we would also be better because we could hint in the setter the use of this enum instead of a string (as of right now). But that would mean inserting that library in the code or use a surrogate where we just use the plain constants in a dedicated class (note how if you use the library when you use `QuickDefinitionMode::MOCK_AT_LEAST_ONCE()` you gey back a object of type `QuickDefinitionMode` not a string, while if we do it with our own class without magic we could just use the constant string)

So our options are:
- we're cool with no being expressive in the API, we can use 2 methods and document it and that's it.
- we introduce the enum for this setting.

Please let me know how you want to proceed and I'll do the changes.

thanks.

davedevelopment:
I keep getting lost between the issue (#1056) and this PR...

Regarding the API and it's extensibility, this is a very narrow use case, so I wouldn't worry about it. Mockery has been going a long time and you're the first to request this, so I'm not overly concerned about making the feature extensible going forward. 

The names do look messy, maybe introduce a small object to wrap the quick definition settings?

``` php
Mockery::configuration()
    ->quickDefinitions()
    ->shouldBeCalledAtLeastOnce(); // pass a boolean to set, no args to get
```

danydev:
@davedevelopment just performed the changes!

davedevelopment:
> I'm not a fan of methods that change semantic based on values of their parameters (i.e. "I'm a getter unless you pass me a value and then I'm setter"). But that's fine for me if that's fine for you.

Generally I agree, but I tend to let it slide for what I would consider to be PHP's lack of (reasonable) support for the Uniform access principle.

Will give @robertbasic a day to object, but I'm happy to merge.

davedevelopment:
@danydev one last request, please could you add a quick note to the CHANGELOG.md file?

Thanks

danydev:
@davedevelopment done

PR #1064, opened 2020-05-25
ExpectationInterface extends Throwable

GrahamCampbell:
Closes https://github.com/mockery/mockery/issues/991.

### DIFF
--- a/library/Mockery/ExpectationInterface.php
+++ b/library/Mockery/ExpectationInterface.php
@@ -21,7 +21,7 @@
 
 namespace Mockery;
 
-interface ExpectationInterface extends Throwable
+interface ExpectationInterface extends \Throwable
 {
     /**
      * @return int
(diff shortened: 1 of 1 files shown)


library/Mockery/ExpectationInterface.php
```diff
@@ -21,7 +21,7 @@
 
 namespace Mockery;
 
-interface ExpectationInterface
+interface ExpectationInterface extends Throwable
```
localheinz, line 24:
```suggestion
interface ExpectationInterface extends \Throwable
```

GrahamCampbell:
Totally wrong... 😆

davedevelopment:
I did wonder :joy:

GrahamCampbell:
Misread it as Exception, but it is Expectation. :trollface:

PR #1075, opened 2020-07-07
[1.3] Support iterable and removed uneeded check

GrahamCampbell:
TODO:
- [x] Implementation
- [x] Tests

---

Closes #869.

### DIFF
--- a/library/Mockery/Mock.php
+++ b/library/Mockery/Mock.php
@@ -724,7 +724,10 @@ class Mock implements MockInterface
             case 'int':    return 0;
             case 'float':  return 0.0;
             case 'bool':   return false;
-            case 'array':  return [];
+
+            case 'array':
+            case 'iterable':
+                return [];
 
             case 'callable':
             case '\Closure':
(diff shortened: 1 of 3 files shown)


library/Mockery/Mock.php
```diff
@@ -740,9 +743,7 @@ public function mockery_returnValueForMethod($name)
                 return null;
 
             case 'object':
-                if (\PHP_VERSION_ID >= 70200) {
```
GrahamCampbell, line 743: This is safe to do because in PHP less than 7.2, `object` will be parsed to `\object` by my new code.

library/Mockery/Mock.php
```diff
@@ -724,7 +724,10 @@ public function mockery_returnValueForMethod($name)
             case 'int':    return 0;
             case 'float':  return 0.0;
             case 'bool':   return false;
-            case 'array':  return [];
+
+            case 'array':
+            case 'iterable':
```
GrahamCampbell, line 729: This is safe because in PHP less than 7.1, iterable will be parsed to \iterable by my new code.

Issue #1084, opened 2020-07-16
Const array throws "Array to string conversion" in ConstantsPass


MPC-khajiit:
Hello,

There is limitation due to passing constans through constantsMap. When you'll try to pass an array to constans map it will throw exception. `sprintf` will try to convert `array` to `string`. 

ex. 

```php
\Mockery::getConfiguration()->setConstantsMap([
    'ClassWithConstants' => [
        'DEFAULTS' => [
            'foo' => 10, 
            'bar' => 20
        ]
    ]
]);

$mock = \Mockery::mock('overload:ClassWithConstants');
```

will throw

```php
ErrorException : Array to string conversion
vendor/mockery/mockery/library/Mockery/Generator/StringManipulation/Pass/ConstantsPass.php:24
vendor/mockery/mockery/library/Mockery/Generator/StringManipulationGenerator.php:79
vendor/mockery/mockery/library/Mockery/Generator/CachingGenerator.php:40
vendor/mockery/mockery/library/Mockery/Container.php:215
vendor/mockery/mockery/library/Mockery.php:117
```

GrahamCampbell:
Don't worry about the PHP 8 test failures. They are due to a very recent change (after my PHP 8 fixes went in), and there is a bug in PHPUnit which won't be released until August with PHPUnit 9.3, so until then, the PHP 8.0 tests will fail. I think that's fine. We can just wait it out. I don't think any mockery releases are planned in the time between now and then.

PR #1085, opened 2020-07-16
Fix: array to string conversion in ConstantsPass [master]

MPC-khajiit:
Fix for #1084

### DIFF
--- a/library/Mockery/Generator/StringManipulation/Pass/ConstantsPass.php
+++ b/library/Mockery/Generator/StringManipulation/Pass/ConstantsPass.php
@@ -21,7 +21,7 @@ class ConstantsPass implements Pass
 
         $constantsCode = '';
         foreach ($cm as $constant => $value) {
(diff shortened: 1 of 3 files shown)


GrahamCampbell:
Is this bug in 1.3.x too? I think we should fix everything which can be reasonable fixed on there, and merge upwards.

MPC-khajiit:
@GrahamCampbell in 1.3.x and 1.2.x too.

GrahamCampbell:
Don't worry about 1.2.x, but if you could target 1.3.x, assuming @davedevelopment is happy with that, I think that'd be best. :)

GrahamCampbell:
I\m a bit confused. This PR has actually already been applied to the code. Why is it being sent again?

GrahamCampbell:
@davedevelopment This can be closed I think?

MPC-khajiit:
@GrahamCampbell  @davedevelopment i'm closing it because previously i prepared two PR (for master and for 1.3 aka #1086) and was merged also into master.

GrahamCampbell:
👍

Issue #1083, opened 2020-07-09
Full PHP 8.0 Support


GrahamCampbell:
PHP 8.0 has not stopped making changes yet, but when they do, we can assess the full situation.

Current PHP 8.0 state:

- [x] Support for all PHP 7.4 syntax on PHP 8.0 (done as of 1.3.2 and 1.4.1)
- [x] Support for union types without null (done as of 1.3.2 and 1.4.1)
- [ ] Support for union types with null (currently broken*)
- [ ] What to do with `Mockery\Generator\Parameter::getClass()` (do we deprecate this function**)
- [ ] Support for the `static` return type hint (I was aware of this before, but possible solutions need to be discussed)
- [ ] Support for other language features (have we missed anything, is there more to come before the 8.0.0RC releases?)

\* I just discovered this an hour ago - no big deal though - as long as it's fixed by November.
\** I don't think this is used anywhere? In my last PR, I updated that function so it basically returns null for union typed cases.

remicollet:
Notice, test suite is failing with 8.0.0RC1

`PHP Fatal error:  Namespace declaration statement has to be the very first statement or after any declare call in the script in /dev/shm/BUILDROOT/php-mockery-1.4.2-1.fc31.remi.x86_64/usr/share/php/Mockery1/Loader/EvalLoader.php(34) : eval()'d code on line 21`

This may looks like a regression, reported as https://bugs.php.net/bug.php?id=80225

a quick fix could be 
```
-        eval("?>" . $definition->getCode());
+        eval(substr($definition->getCode(), 5));

```
Probably worth waiting for some inputs on this upstream issue

GrahamCampbell:
Thanks for letting us know @remicollet. :)

remicollet:
This in 8.0.0RC2 ;)

GrahamCampbell:
Fix commit: https://github.com/php/php-src/commit/f9b7609d17694c5b5d5bba6321b27944220566be.

raveren:
whooo an hour wasted debugging 

```
ParseError : syntax error, unexpected token "static", expecting identifier
 /private/var/folders/4r/nyfsnst15kz019tl_4xmfml80000gn/T/MockeryYx5sWT:1
```

Turns out Mockery does not support `static` return type on methods. 

Hear that, google? **Mockery does not support `static` return type!**

[...]

PR #1088, opened 2020-08-03
[1.3] Improved PHP 8.0 support

GrahamCampbell:
This PR addresses some issues from the previous PR, some of which were caused by additional changes made to PHP 8.0, after the other PR was done. This doesn't address automatic creation of return values for methods typed as `static`. I am not sure if we even want to support this...

---

Depends on https://github.com/mockery/mockery/pull/1087. Related to https://github.com/mockery/mockery/issues/1083.

### DIFF
--- a/.styleci.yml
+++ b/.styleci.yml
@@ -9,4 +9,5 @@ disabled:
 
 finder:
   not-name:
+    - Php80LanguageFeaturesTest.php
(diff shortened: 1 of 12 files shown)


GrahamCampbell:
@davedevelopment Ready for review, now. :)

PR #1096, opened 2020-10-17
Update type hints

crabbly:
Just a small update to a few type hints to prevent IDEs from flagging. Below is a screenshot using PhpStorm.

Thank you for this awesome lib! 🚀 

mockery

### DIFF
--- a/library/Mockery/Matcher/MatcherAbstract.php
+++ b/library/Mockery/Matcher/MatcherAbstract.php
@@ -25,7 +25,7 @@ abstract class MatcherAbstract
     /**
      * The expected value (or part thereof)
      *
-     * @var mixed
+     * @var array|mixed[]
      */
     protected $_expected = null;
 
--- a/library/Mockery/Matcher/MatcherAbstract.php
+++ b/library/Mockery/Matcher/MatcherAbstract.php
@@ -25,7 +25,7 @@ abstract class MatcherAbstract
     /**
      * The expected value (or part thereof)
      *
-     * @var array|mixed[]
+     * @var mixed
      */
     protected $_expected = null;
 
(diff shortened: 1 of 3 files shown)


library/Mockery/Matcher/MatcherAbstract.php
```diff
@@ -25,7 +25,7 @@ abstract class MatcherAbstract
     /**
      * The expected value (or part thereof)
      *
-     * @var mixed
+     * @var array|mixed[]
      */
     protected $_expected = null;
```
crabbly, line 30: This is the opposite from the other two changes because by the time we assign the constructor input to this property, php have already transformed a single or multiple arguments into an array.

library/Mockery/Matcher/MatcherAbstract.php
```diff
@@ -25,7 +25,7 @@ abstract class MatcherAbstract
     /**
      * The expected value (or part thereof)
      *
-     * @var mixed
+     * @var array|mixed[]
```
GrahamCampbell, line 28: This change is wrong, and there are other places that need to also be corrected, like the constructor phpdoc.

GrahamCampbell (changes requested):

library/Mockery/Matcher/MatcherAbstract.php
crabbly: Hey @GrahamCampbell, sure I'll push a commit removing it. The purpose for the MR was really for the `andReturn` type hint.

However, could you please explain why? I'm wondering what I'm missing.

The phpdoc for the constructor is to provide type hints for consumers of that class, it has nothing to do with this property. At object instantiation, the constructor arguments are passed into an array, and only then we assign this new array to the local (protected) property.

https://www.php.net/manual/en/functions.arguments.php#functions.variable-arg-list.new

crabbly: This is indeed wrong. I mixed this constructor with another class. Nice catch. Fixed!

PR #1097, opened 2020-10-17
Allow shouldIgnoreMissing() to behave in a recursive fashion

AJenbo:
When setting up lazy mocks I often run into issues where the test just breaks at the secound link and I end up having to explicity set expectations for the functions any way. This PR aims to solve that by allowing for `shouldIgnoreMissing()` to behave in a recursive fashion.

code:
```php
class A {
    function now(): DateTime {
        return new DateTime();
    }
}
class Factory {
    function makeA(): A {
        return new A();
    }
}
class Aubject {
    $factory;
    function __construct($factory) {
        $this->factory = $factory;
    }
    function testMe() {
        return $this->factory->makeA()->now()->format('c');
    }
}
```

Before:
```php
$mockedFactory = Mockery::spy(Factory::class):
$aMock->allows('now')->andReturn(Mockery::spy(DateTime::class));
$mockedFactory->allows('makeA')->andReturn($aMock);

$subject = new Aubject($mockedFactory);
$subject->testMe();
```

After:
```php
$mockedFactory = Mockery::mock(Factory::class)->shouldIgnoreMissing(null, true):

$subject = new Aubject($mockedFactory);
$subject->testMe();
```

### DIFF
--- a/library/Mockery/Mock.php
+++ b/library/Mockery/Mock.php
@@ -53,6 +53,14 @@ class Mock implements MockInterface
     protected $_mockery_ignoreMissing = false;
 
     /**
+     * Flag to indicate whether we want to set the ignoreMissing flag on
+     * mocks generated form this calls to this one
+     *
+     * @var bool
+     */
+    protected $_mockery_ignoreMissingRecursive = false;
+
+    /**
      * Flag to indicate whether we can defer method calls missing from our
      * expectations
      *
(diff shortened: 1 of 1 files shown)


library/Mockery/Mock.php
```diff
@@ -308,11 +316,13 @@ public function shouldAllowMockingMethod($method)
     /**
      * Set mock to ignore unexpected methods and return Undefined class
      * @param mixed $returnValue the default return value for calls to missing functions on this mock
+     * @param bool $recursive Speficy if returned mocks should also have shouldIgnoreMissing set
```
GrahamCampbell, line 319: Speficy? ;)

GrahamCampbell (changes requested):

library/Mockery/Mock.php
AJenbo:
```suggestion
     * @param bool $recursive Specify if returned mocks should also have shouldIgnoreMissing set
```

davedevelopment:
Thank you

PR #1104, opened 2020-11-25
fix shouldHaveReceived $args param typehint

e-zannelli:
Current type hint suggests that $args can only be null, and static analysis tools complain when you call the method with something else.

Add types that you can pass to withArgs.

Thanks


GrahamCampbell:
This should probably be fixed on the 1.3.x branch.

e-zannelli:
rebased

davedevelopment:
Thank you

PR #1106, opened 2020-12-06
Fix crash on a union type including null

### LINKED ISSUE #1105: Union type like "array|string|null" or "string|int|null" crashes Mockery
The following code snippet crashes when the code generated by Mockery is evaluated.

```php
class Foo {
    public function bar(array|string|null $v) {}
}

Mockery::mock(Foo::class);
```

In this situation, the generated code looks like this:
```php
public function bar(array|string|null|null $v) {
    $argc = func_num_args();
    $argv = func_get_args();
    $ret = $this->_mockery_handleMethodCall(__FUNCTION__, $argv);
    return $ret;
}
```

I found this issue because of a mock of a class provided by a PHP extension, so I took a quick look at the other classes declared in my environment. At least the following classes are affected:
* `MultipleIterator::class`
* `DOMXPath::class`
* `IntlDateFormatter::class`
* `Phar::class`
* `PharData::class`
* `SoapFault::class`
* `SoapHeader::class`
* `XSLTProcessor::class`
* `ZipArchive::class`

It seems the support introduced by #1088 needs to be improved a bit to manage this case. I will try to take a look later this week.

LeSuisse:
This change prevents the generation of type hints with 2 "null".

For example, "string|array|null" should generate the code "string|array|null" instead of "string|array|null|null" to avoid a PHP fatal error "Duplicate type null is redundant".

Closes #1105.

### DIFF
--- a/library/Mockery/Reflector.php
+++ b/library/Mockery/Reflector.php
@@ -192,9 +192,11 @@ class Reflector
     {
         // PHP 8 union types can be recursively processed
         if ($type instanceof \ReflectionUnionType) {
-            return \implode('|', \array_map(function (\ReflectionType $type) use ($declaringClass) {
-                return self::typeToString($type, $declaringClass);
-            }, $type->getTypes()));
+            return \implode('|', \array_filter(\array_map(function (\ReflectionType $type) use ($declaringClass) {
+                $typeHint = self::typeToString($type, $declaringClass);
+
+                return $typeHint === 'null' ? null : $typeHint;
+            }, $type->getTypes())));
         }
 
         // PHP 7.0 doesn't have named types, but 7.1+ does
--- a/tests/PHP80/Php80LanguageFeaturesTest.php
+++ b/tests/PHP80/Php80LanguageFeaturesTest.php
@@ -30,6 +30,15 @@ class Php80LanguageFeaturesTest extends MockeryTestCase
     }
 
     /** @test */
+    public function it_can_mock_a_class_with_a_union_argument_type_hint_including_null()
+    {
+        $mock = mock(ArgumentUnionTypeHintWithNull::class);
+        $mock->allows()->foo(null);
+
+        $mock->foo(null);
+    }
+
+    /** @test */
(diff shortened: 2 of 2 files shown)


GrahamCampbell:
Does that test pass without the code change. When I originally wrote this code `string|null` would not show up as a union type, and PHP would pretend it was `?string`. I wonder if the bug you were experiencing is a bug with union types in core code, and not userland functions?

LeSuisse:
No it does not, see https://travis-ci.com/github/LeSuisse/mockery/jobs/452045803 which correspond to https://github.com/LeSuisse/mockery/commit/d328ed672ff6b9fc5be01848371d6e7750ab1d32 (the test without the code change).
I detected the issue by mocking a core class but it seems userland is also affected.

LeSuisse:
The CI failures with `DEPS=lowest` on PHP 7.3 and PHP 7.4 seems to be unrelated to this change (I did not confirm it but it looks like a side effect of the release of Xdebug 3.0).

GrahamCampbell:
NB This PR should target the 1.3 branch. PHPUnit issues can be fixed by replacing `^5.7.10|^6.5|^7.5|^8.5|^9.3` with the same thing, only with each major version min changed to very latest in that series.

LeSuisse:
The PR now targets the 1.3 branch.

Not exactly sure to understand how to fix the PHPUnit/Xdebug 3.0 issues on it since we need PHPUnit ^5 .7 to run the tests on PHP 5.6 and this version also covers PHP ^7.0.

GrahamCampbell:
Hmm, mockery should probably just drop it's prefer-lowest tests. They don't add any value, since Mockery has basically no dependencies.

davedevelopment:
This looks good to me and I'm happy to drop the prefer-lowest tests, to be honest that all confuses me anyway.

LeSuisse:
I opened #1107 to drop the prefer-lowest tests. I will rebase this PR once the other is merged :) .

davedevelopment:
Sorry for the delay, just tagged 1.4.3 and 1.3.4

PR #1107, opened 2020-12-08
Stop testing with '--prefer-lowest'

LeSuisse:
This solve issues with PHPUnit and Xdebug 3.0 when running tests for PHP 7.3 and 7.4.

mockery has basically no dependencies so the part of the matrix testing with '--prefer-lowest' brings little/no value.

See the discussion in https://github.com/mockery/mockery/pull/1106#issuecomment-739831499

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,49 +3,12 @@ language: php
 matrix:
   include:
     - php: 5.6
-      env:
-        - DEPS=lowest
-    - php: 5.6
-      env:
-        - DEPS=latest
     - php: 7.0
-      env:
-        - DEPS=lowest
-    - php: 7.0
-      env:
-        - DEPS=latest
     - php: 7.1
-      env:
-        - DEPS=lowest
-    - php: 7.1
-      env:
-        - DEPS=latest
     - php: 7.2
-      env:
-        - DEPS=lowest
-    - php: 7.2
-      env:
-        - DEPS=latest
     - php: 7.3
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -3,49 +3,12 @@ language: php
...
-      env:
-        - DEPS=lowest
     - php: 7.4
-      env:
-        - DEPS=latest
-    - php: nightly
-      env:
-        - DEPS=lowest
-        - COMPOSER_FLAGS="--ignore-platform-reqs"
     - php: nightly
```
GrahamCampbell, line 11: This should be changed to `8.0`.

GrahamCampbell (changes requested):

.travis.yml
LeSuisse: Thanks I do not follow closely what's going on in the Travis CI world these days, I was not aware the 8.0 was available.

LeSuisse:
Would you like that I open the equivalent PR for the `master` branch?

davedevelopment:
@LeSuisse Sure, will save me a job thank you  :+1:

GrahamCampbell:
Usually what I do is an actual merge 1.3 -> master, so the commit history is retained. We can probably hold off merging upwards until the other upcoming 1.3 PR lands.

davedevelopment:
Sorry, forgot we had the other one, all merged.

PR #1109, opened 2020-12-11
Build API docs with phpDocumentor 3.0.0 final instead of RC

LeSuisse:
phpDocumentor cache has been added to the .gitignore so it does not end up committed into the repo/a PR by mistake.


davedevelopment:
Thank you!

PR #1113, opened 2021-01-14
Fix Mockery::fetchMock() without initialized container

PhrozenByte:
Calling `Mockery::fetchMock()` first doesn't make much sense and likely indicates a logic error, but Mockery should nevertheless throw `Call to a member function fetchMock() on null` errors.

Maybe we should throw a `LogicException` instead when no container was initialized yet (just like `Mockery::self()`)? If no container was initialized yet, `Mockery::fetchMock()` will always return `null`, just like if the mock in question simply doesn't exist (yet). Other opinions?

### DIFF
--- a/library/Mockery.php
+++ b/library/Mockery.php
@@ -214,7 +214,7 @@ class Mockery
      */
     public static function fetchMock($name)
     {
-        return self::$_container->fetchMock($name);
+        return self::getContainer()->fetchMock($name);
     }
 
     /**


GrahamCampbell:
This should probably be fixed on the 1.3.x branch.

PhrozenByte:
Are you guys working in the `1.3` branch and merging changes back to `master` (i.e. I should create a PR on the `1.3` branch and close this one), or can this be merged into `master` and merged into `1.3`? Or do you prefer separate PRs with different commits?

GrahamCampbell:
I would say just send to 1.3. I can handle any 1.3 -> master merges if needed. :)

GrahamCampbell:
The reason for this is just that 1.4.x doesn't support PHP 7.1 and 7.2, so it makes sense to land bug fixes in the 1.3.x series, for the rest of the year, at least.

PhrozenByte:
Here you go #1114 :+1:

PR #1126, opened 2021-05-25
update meta to properly infer type of string parameters

isfedorov:
After the fix of https://youtrack.jetbrains.com/issue/WI-60544 type() meta directive won't treat string literals contents as types anymore, prefered way is to use (map([""=>"$0"])). This PR updates meta file according to preferred way.

### DIFF
--- a/.phpstorm.meta.php
+++ b/.phpstorm.meta.php
@@ -2,10 +2,10 @@
 
 namespace PHPSTORM_META;
 
-override(\Mockery::mock(0), map(["" => "$0"]));
-override(\Mockery::spy(0), map(["" => "$0"]));
-override(\Mockery::namedMock(0), map(["" => "$0"]));
-override(\Mockery::instanceMock(0), map(["" => "$0"]));
-override(\mock(0), map(["" => "$0"]));
-override(\spy(0), map(["" => "$0"]));
-override(\namedMock(0), map(["" => "$0"]));
+override(\Mockery::mock(0), map(["" => "@"]));
+override(\Mockery::spy(0), map(["" => "@"]));
+override(\Mockery::namedMock(0), map(["" => "@"]));
+override(\Mockery::instanceMock(0), map(["" => "@"]));
+override(\mock(0), map(["" => "@"]));
+override(\spy(0), map(["" => "@"]));
+override(\namedMock(0), map(["" => "@"]));
--- a/.phpstorm.meta.php
+++ b/.phpstorm.meta.php
@@ -2,10 +2,10 @@
 
 namespace PHPSTORM_META;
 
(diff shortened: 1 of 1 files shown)


davedevelopment:
I don't know much about this, so :+1: from me without any real authority...

KapitanOczywisty:
@isfedorov
1. It is better to use `@` instead of `$0`, since latter may not work in tools like intelephense
2. There should be also added `MockInterface` since it is included in resulting object
```php
override(
    \Mockery::mock(0),
    map([""=>"@&\Mockery\MockInterface"])
);
```

isfedorov:
@KapitanOczywisty 
>There should be also added MockInterface since it is included in resulting object

Seems there is no such need because `Mockery\MockInterface` is received from 
```
/**
     * Static shortcut to \Mockery\Container::mock().
     *
     * @param mixed ...$args
     *
     * @return \Mockery\MockInterface|\Mockery\LegacyMockInterface
     */
    public static function mock(...$args)
    {
        return call_user_func_array(array(self::getContainer(), 'mock'), $args);
    }
```
So I've changed only `$0` to `@`
Screenshot 2021-07-16 at 19 19 58

KapitanOczywisty:
> Seems there is no such need because `Mockery\MockInterface` is received from

Intelephense is treating `@return` as fallback and it is not merging definitnions, I don't have access to PHPStorm so I was unable to verify how it suppose to work. I'd add interface anyway for that reason, just because it could be a while before it will be fixed in intelephense.

mad-briller:
any plans on merging this? would be a great improvement to the usability of mockery with modern tools

davedevelopment:
@mad-briller I'm happy to merge, but it didn't seem like there was a consensus on how it should be. Will give it a few days and if no further objections or justifications come in, I'll merge.

KapitanOczywisty:
@davedevelopment I'd still use:

```php
override(\Mockery::mock(0), map(["" => "@&\Mockery\MockInterface"]));
override(\Mockery::spy(0), map(["" => "@&\Mockery\MockInterface"]));
override(\Mockery::namedMock(0), map(["" => "@&\Mockery\MockInterface"]));
override(\Mockery::instanceMock(0), map(["" => "@&\Mockery\MockInterface"]));
override(\mock(0), map(["" => "@&\Mockery\MockInterface"]));
override(\spy(0), map(["" => "@&\Mockery\MockInterface"]));
override(\namedMock(0), map(["" => "@&\Mockery\MockInterface"]));
```

PHPStorm will work in both cases, but tools like intelephense needs explicit type intersection.

In any case this is a valid PR and can be merged.

davedevelopment:
:+1:

PR #1128, opened 2021-06-24
updated reserved methods in docs

unstoppablecarl:
Updated the reserved methods in the docs. I am curious what `shouldBeStrict()` was?


davedevelopment:
It was something to do with the recorder feature, which the last remnants of were removed in 916a93e70e2164e1ae04ec0442f742ceeb265df0

It was something to do with "recording" expectations, rather than declaring them. So your test code acted out what it expected to happen as a recording, then set the production code going which should replicate it.

PR #1130, opened 2021-06-26
Fixes for PHP 8.1

jrmajor:
- Avoid passing null to non-nullable parameters of internal functions
- Skip `Serializable` test on PHP <8.1
- Add support for tentative return types

### DIFF
--- a/library/Mockery/Mock.php
+++ b/library/Mockery/Mock.php
@@ -803,7 +803,7 @@ class Mock implements MockInterface
             throw new BadMethodCallException(
                 'Static method ' . $associatedRealObject->mockery_getName() . '::' . $method
                 . '() does not exist on this mock object',
-                null,
+                0,
                 $e
             );
         }
--- a/library/Mockery/Reflector.php
+++ b/library/Mockery/Reflector.php
@@ -84,11 +84,20 @@ class Reflector
     public static function getReturnType(\ReflectionMethod $method, $withoutNullable = false)
     {
         // Strip all return types for HHVM and skip PHP 5.
-        if (method_exists($method, 'getReturnTypeText') || \PHP_VERSION_ID < 70000 || !$method->hasReturnType()) {
+        if (method_exists($method, 'getReturnTypeText') || \PHP_VERSION_ID < 70000) {
             return null;
         }
 
         $type = $method->getReturnType();
+
+        if (is_null($type) && method_exists($method, 'getTentativeReturnType')) {
+            $type = $method->getTentativeReturnType();
+        }
+
+        if (is_null($type)) {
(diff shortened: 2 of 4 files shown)


GrahamCampbell:
Can you target the 1.3 branch please, so Laravel can get these fixes.

jrmajor:
@GrahamCampbell Sorry, I wasn't aware that Mockery doesn't follow semantic versioning. I've never changed target branch in PR, so I'm not sure how to do it correctly, will it work if I just change it via GitHub interface?

I see Laravel 8 requires `^1.4` and 6 requires `~1.3.3|^1.4.2`, so both current and LTS versions should be able to get this fixes. Why is it needed to target 1.3?

GrahamCampbell:
Mockey does follow semantic versioning, but the 1.4.x line has cranked the min PHP version. Laravel needs the 1.3.x line to continue to be maintained for the rest of 2021 (and I'm happy to do any and all additional work to make that happen). :)

jrmajor:
> Mockey does follow semantic versioning, but the 1.4.x line has cranked the min PHP version.

OK, I see. The strange version constraint made me think so. Wouldn't `~1.3.3` behave the same as `^1.3.3`?

And please let me know what I need to do to change the target branch. I'm also happy to do this work, but in this case I'm not sure how :)

GrahamCampbell:
> Wouldn't ~1.3.3 behave the same as ^1.3.3?

No. `^` is a semver operator. `~` pins the major and minor in your example.

jrmajor:
@GrahamCampbell Thanks for your anwsers :) I've changed the target to 1.3, but I can't run tests (`Test directory "tests/PHP56" not found`).

GrahamCampbell:
Looks like we need to move to GH actions. I'll create an issue and do it soon.

GrahamCampbell (approved):

davedevelopment:
Merged. I'm a bit short on time right now, summer holidays, kids etc, so I've not given it a proper good looking at, trusting @GrahamCampbell :+1:

jrmajor:
@davedevelopment Thank you!

`Mockery\Reflector::getReturnType()` is different in 1.3 and 1.4, so you may use version from b4da5440991f6c2f517e2879b5bb15772a2a8796 to resolve this conflict when merging `1.3` to `master`.

GrahamCampbell:
That change to 1.4.x should probably also be backported to 1.3.x.

jrmajor:
@GrahamCampbell The commit I mentioned is the version for 1.4, which on your request I later backported to 1.3. It had conflicts in this method, so this version may come in handy when merging this PR to 1.4.

PR #1133, opened 2021-07-06
Add method that allows defining a set of arguments the mock should yield

### LINKED ISSUE #520: No way to mock methods returning generators
There should be a way to mock methods that return generators, i.e. `MockInterface::andYield($foo, $bar, ...)`.

Blacksmoke16:
Resolves #520

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Change Log
 
+## 1.4.4 (TBC)
+* Add method that allows defining a set of arguments the mock should yield #1133
+
 ## 1.4.3 (2021-02-24)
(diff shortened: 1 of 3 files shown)


davedevelopment:
Please could you add a note to the changelog and I'll get this merged.

Blacksmoke16:
@davedevelopment Do you want it to be a specific version/date or with like question marks as placeholders for now?

davedevelopment:
Sorry, didn't realise we didn't have a heading yet. Yes and no, something like this would be great :)

``` markdown
# 1.4.4 (TBC)

- Summary of your change
```

Blacksmoke16:
@davedevelopment Done.

PR #1146, opened 2021-09-16
Override default call count expectations via `expects()`

### LINKED ISSUE #1138: expects('foo')->atLeast()->once() errors on two calls
```php
$mock = \Mockery::mock(Foo::class)->expects('foo')->atLeast()->once();
$mock->foo(); 
$mock->foo(); 
```

>Mockery\Exception\InvalidCountException : Method advance(<Any Arguments>) from Mockery_1_Foo should be called
 exactly 1 times but called 2 times.

ghostwriter:
**By default `expects()` sets up an expectation that the method should be called once and once only.** 

http://docs.mockery.io/en/stable/reference/alternative_should_receive_syntax.html#expects
http://docs.mockery.io/en/stable/reference/expectations.html#declaring-call-count-expectations

If accepted, this example code will run as expected.
```php
$this->mock->expects('foo')->atLeast()->once();

// and or

$this->mock->expects()->foo()->atMost()->once();
```

- Add failing test
- Implement the bug fix 
- Resolves #1138

### DIFF
--- a/library/Mockery/Expectation.php
+++ b/library/Mockery/Expectation.php
@@ -728,7 +728,12 @@ class Expectation implements ExpectationInterface
             throw new \InvalidArgumentException('The passed Times limit should be an integer value');
         }
         $this->_countValidators[$this->_countValidatorClass] = new $this->_countValidatorClass($this, $limit);
-        $this->_countValidatorClass = 'Mockery\CountValidator\Exact';
+
+        if('Mockery\CountValidator\Exact' !== $this->_countValidatorClass){
+            $this->_countValidatorClass = 'Mockery\CountValidator\Exact';
+            unset($this->_countValidators[$this->_countValidatorClass]);
+        }
+
         return $this;
     }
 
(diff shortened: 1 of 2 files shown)


Bilge:
@davedevelopment How does this look to you?

freiondrej-lmc:
@davedevelopment it would be amazing if you could tag this, we've just encountered this behavior in our test suite. It's great that it's been fixed already, but we don't want to require dev-master

davedevelopment:
Sorry for the delay. I'll see if I can get around to tagging sometime soon.

Bilge:
Bro, what? You just type `git tag ver` and `git push --tag`. You literally spent more time typing that comment than just pushing a tag.

davedevelopment:
I usually spend a bit of time reviewing what has changed and updating the CHANGELOG etc. It doesn't take long, but I don't have that much time to contribute and I'd rather not rush. 

Bilge has annoyed me enough times, he's rarely constructive so I've blocked him.

PR #1156, opened 2022-01-11
Mock methods with `mixed` return type

### LINKED ISSUE #1149: Cannot use "mixed" as class name
We're getting the following error when running our PHPUnit suite:
> Cannot use 'mixed' as class name as it is reserved in /tmp/MockerykqgHQz on line 1

PHP version: 8.0.8
Mockery version: 1.4.4
PHPUnit version: 9.5.10

---

The contents of the generated file:
```php
<?php class mixed {} 
```

ghostwriter:
This patch resolves #1149, via the following changes:

 - Add tests and fixtures
 - Mock methods with `mixed` return type.

``` php
class MockingMethodsWithMixedReturnTypeTest extends MockeryTestCase
{
    public function testMockingMixedReturnType()
    {
        $mock = \Mockery::mock(MyInterface::class);
        $mock->shouldReceive("foo->bar")->andReturn("bar");
        $this->assertSame('bar', $mock->foo()->bar());
    }
}

interface MyInterface {
    public function getFoo(): mixed;
}
```

### DIFF
--- a/library/Mockery.php
+++ b/library/Mockery.php
@@ -886,7 +886,7 @@ class Mockery
             $parRefMethod = $parRef->getMethod($method);
             $parRefMethodRetType = Reflector::getReturnType($parRefMethod, true);
 
-            if ($parRefMethodRetType !== null) {
+            if ($parRefMethodRetType !== null && $parRefMethodRetType !== 'mixed') {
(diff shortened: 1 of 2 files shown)


davedevelopment:
Thank you!

PR #1157, opened 2022-01-11
Mock methods with `static` return types

### LINKED ISSUE #1155: Attempting to mock a method with `static` return type in PHP 8 results in ParseError
[PHP 8.0 introduced the `static` return type.](https://www.php.net/manual/en/language.types.declarations.php#language.types.declarations.static)

When using Mockery to mock a method with a return type hint of `static`, PHP generates a `ParseError`, with a message similar to the following:

```
ParseError : syntax error, unexpected token "static", expecting identifier /private/var/folders/jt/dm973jds5b33ffvl0x2b8f940000gn/T/MockeryYyJCuE:1
```

The generated file contains the following:

```php
<?php class static {} 
```

#1149 describes a similar issue with the `mixed` return type.

## To Reproduce

💡 **NOTE**: To help, I've created a repository so you can quickly reproduce the issue: https://github.com/ramsey/mockery-static-return-type-bug

Create a class with a method on it that has a `static` return type hint:

``` php
<?php

namespace Foo;

class Foo
{
    public function setFoo(string $value): static
    {
        return $this;
    }
}
```

Create a PHPUnit test that attempts to create a Mockery mock of the class and method:

``` php
<?php

namespace Foo\Tests;

use Foo\Foo;
use Mockery;
use PHPUnit\Framework\TestCase;

class FooTest extends TestCase
{
    public function testFoo(): void
    {
        $fooMock = Mockery::mock(Foo::class);
        $fooMock->allows('setFoo');

        $fooMock->setFoo('some value');

        $this->assertTrue(true);
    }
}
```

Run the test on PHP 8.0 or 8.1.

ghostwriter:
This patch resolves #1155, via the following changes:

 - Add tests and fixtures
 - Mock methods with `static` return type.

``` php
class Foo
{
    public function getFoo(): static
    {
        return $this;
    }
}
```

### DIFF
--- a/library/Mockery/Mock.php
+++ b/library/Mockery/Mock.php
@@ -747,6 +747,9 @@ class Mock implements MockInterface
             case 'void':
                 return null;
 
+            case 'static':
+                return $this;
+                
             case 'object':
                 $mock = \Mockery::mock();
                 if ($this->_mockery_ignoreMissingRecursive) {
--- a/library/Mockery/Mock.php
+++ b/library/Mockery/Mock.php
@@ -749,7 +749,7 @@ class Mock implements MockInterface
 
             case 'static':
                 return $this;
-                
+
             case 'object':
                 $mock = \Mockery::mock();
                 if ($this->_mockery_ignoreMissingRecursive) {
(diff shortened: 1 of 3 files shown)


davedevelopment:
Thank you!

ramsey:
Thanks, @ghostwriter!

ruudk:
Thanks for this. Would be great to have this tagged 🙏

PR #1160, opened 2022-01-11
Mock classes with `new` in initializers on PHP 8.1

### LINKED ISSUE #1154: PHP 8.1 new in constructor initializers
Mockery does not support `new` in constructor initializers. That is, the following cannot be mocked without an error.

```php
class MyController {
    public function __construct(
        private Logger $logger = new NullLogger(),
    ) {}
}
```

Mocking this will generate an error similar to the following:

>PHP Fatal error:  Constant expression contains invalid operations in mockery\mockery\library\Mockery\Loader\EvalLoader.php(34) : eval()'d code on line 974

ghostwriter:
This patch resolves #1154, via the following changes:

 - Add tests and fixtures 
 - Mock classes with `new` in initializers

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -5,19 +5,19 @@ on:
   pull_request:
 
 jobs:
-  php7:
-    name: PHP ${{ matrix.php }}
-    runs-on: ubuntu-20.04
+  phpunit:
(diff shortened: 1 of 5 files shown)


Bilge:
Confirmed this patch works for me 👍🏻

davedevelopment:
👍 thank you

PR #1169, opened 2022-03-15
PHP 8.2 | Fix "Use of "parent" in callables is deprecated" notice

### LINKED ISSUE #1168: PHP 8.2: tests error out on "Use of "parent" in callables is deprecated"
### Description

PHP 8.2 is slated to [deprecate partially supported callables](https://wiki.php.net/rfc/deprecate_partially_supported_callables).

From the [RFC](https://wiki.php.net/rfc/deprecate_partially_supported_callables#backward_incompatible_changes):
> Most of the callables deprecated here have a straightforward replacement: "self" should be replaced with self::class, and so on:
> ```
> "self::method"       -> self::class . "::method"
> "parent::method"     -> parent::class . "::method"
> "static::method"     -> static::class . "::method"
> ["self", "method"]   -> [self::class, "method"]
> ["parent", "method"] -> [parent::class, "method"]
> ["static", "method"] -> [static::class, "method"]
> ```
> The new form of these callables is no longer context-dependent. It will refer to the self/parent/static scope of where the callable has been created, rather than where is will be called.

During a preliminary test run for a project against PHP 8.2, I noticed that Mockery is heavily affected by this issue, with over half of the tests in the project I did the test run on erroring out on a `Use of "parent" in callables is deprecated` error.

### How to reproduce

* Clone the https://github.com/jrfnl/bug-report-reproduction-scenarios/ repo
* Check out the `mockery/php-8.2-error-with-partial-callables` branch
* Run `composer install`
* Run `vendor/bin/phpunit` on PHP 8.2

Or examine the results of the workflows runs for this branch: https://github.com/jrfnl/bug-report-reproduction-scenarios/actions/runs/1985112787

### Error

```
There was 1 error:

1) Jrf\Mockery\Scenario\Tests\FooTest::testSplitString with data set "simple test case" ('test,string', array('test', 'string'))
Use of "parent" in callables is deprecated

/home/runner/work/bug-report-reproduction-scenarios/bug-report-reproduction-scenarios/tests/FooTest.php:33
```

jrfnl:
I've searched the code base to see if I could find all affected callbacks. No guarantee I've found them all though, though a test run against PHP 8.2 with the testsuite of this codebase no longer shows this particular error with the patch applied.

Fixes #1168

### DIFF
--- a/library/Mockery/Mock.php
+++ b/library/Mockery/Mock.php
@@ -614,7 +614,7 @@ class Mock implements MockInterface
     public function __isset($name)
     {
         if (false === stripos($name, '_mockery_') && get_parent_class($this) && method_exists(get_parent_class($this), '__isset')) {
-            return call_user_func('parent::__isset', $name);
+            return call_user_func(get_parent_class($this) . '::__isset', $name);
         }
 
         return false;
@@ -637,9 +637,9 @@ class Mock implements MockInterface
     public function mockery_callSubjectMethod($name, array $args)
     {
         if (!method_exists($this, $name) && get_parent_class($this) && method_exists(get_parent_class($this), '__call')) {
-            return call_user_func('parent::__call', $name, $args);
+            return call_user_func(get_parent_class($this) . '::__call', $name, $args);
         }
-        return call_user_func_array('parent::' . $name, $args);
+        return call_user_func_array(get_parent_class($this) . '::' . $name, $args);
     }
 
     /**
@@ -858,7 +858,7 @@ class Mock implements MockInterface
                 // noop - there is no hasPrototype method
             }
 
-            return call_user_func_array("parent::$method", $args);
+            return call_user_func_array(get_parent_class($this) . '::' . $method, $args);
         }
 
         $handler = $this->_mockery_findExpectedMethodHandler($method);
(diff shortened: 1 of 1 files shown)


ghost (approved):

driesvints:
Can we maybe get a release for this? Right now, we can't add passing PHP 8.2 builds to Laravel and some of our first party libraries because of the deprecation notices. This is one of the few things that's still blocking us so would be cool to see this getting a tag! : )

jrfnl:
@driesvints FYI: There are actually still some more PHP 8.2 issues to be fixed in Mockery (or to find work-arounds for). I still need to find the time to write those up properly.

driesvints:
@jrfnl definitely. But right now, the code of this PR is the only thing we encounter ourselves. Is there a chance we can get that released so you can do the remaining ones in an upcoming release? Dave mentioned he wanted to check into this tomorrow: https://twitter.com/davedevelopment/status/1567170288642572296

jrfnl:
> Is there a chance we can get that released so you can do the remaining ones in an upcoming release?

Not my decision (or responsibility). Just like you, I'm just a user of Mockery and I fully support a release with this fix, even if it's only partial PHP 8.2 support, it gets rid of most issues for most code bases using Mockery.

driesvints:
Thanks @jrfnl!

cmanley:
Shouldn't this:
`call_user_func(get_parent_class($this) . '::__isset', $name);`
be written as 
`call_user_func(parent::class . '::__isset', $name);`
?
Because $this doesn't have to be an instance of the current class: it could be a child class in which case you'll end up with infinite recursion.

jrfnl:
> Shouldn't this: `call_user_func(get_parent_class($this) . '::__isset', $name);` be written as `call_user_func(parent::class . '::__isset', $name);` ? Because $this doesn't have to be an instance of the current class: it could be a child class in which case you'll end up with infinite recursion.

@cmanley The call is conditional and the code as per this PR follows the code in the condition:

```php
if (false === stripos($name, '_mockery_') && get_parent_class($this) && method_exists(get_parent_class($this), '__isset')) {
```

PR #1184, opened 2022-06-07
[8.1] Support `never` return type

### LINKED ISSUE #1167: A never-returning function must not return in /vendor/mockery/mockery/library/Mockery/Loader/EvalLoader.php
Hi, I've got this error message when try to mock a class. You can reproduce this by:

```php
trait ResponsesTrait {
    public function notFound(): never
    {
        throw new RuntimeException('Not found');
    }
}
```

```php
class Controller {
    use ResponsesTrait;
    
    public function index(): Response
    {
        $this->notFound();
    }
}
```

And in the test:

```
$controller = Mockery::mock(new Controller())->makePartial();
```

ghostwriter:
This patch resolves #1167, via the following changes:

 - MethodDefinitionPass will not generate a `return` statement for a `never` return typehint.

A function/method that is declared with the `never` return type indicates that it will never `return` a value, and **always** throws an exception or terminates with a `die/exit `call.

### DIFF
--- a/library/Mockery/Generator/StringManipulation/Pass/MethodDefinitionPass.php
+++ b/library/Mockery/Generator/StringManipulation/Pass/MethodDefinitionPass.php
@@ -157,7 +157,7 @@ BODY;
 
         $body .= "\$ret = {$invoke}(__FUNCTION__, \$argv);\n";
 
-        if ($method->getReturnType() !== "void") {
+        if (! in_array($method->getReturnType(), ['never','void'], true)) {
             $body .= "return \$ret;\n";
         }
 
--- a/tests/PHP81/Php81LanguageFeaturesTest.php
+++ b/tests/PHP81/Php81LanguageFeaturesTest.php
@@ -8,6 +8,9 @@ use Mockery\Adapter\Phpunit\MockeryTestCase;
 use ReturnTypeWillChange;
 use RuntimeException;
 use Serializable;
+use function pcntl_fork;
+use function pcntl_waitpid;
+use function pcntl_wexitstatus;
 
 /**
  * @requires PHP 8.1.0-dev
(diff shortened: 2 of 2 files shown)


ghostwriter:
I'm not sure how to go about testing the `die/exit` case. 

Would love some feedback and or guidance.

davedevelopment:
@ghostwriter this is one way, but might be too flaky to bother

``` diff
diff --git tests/PHP81/Php81LanguageFeaturesTest.php tests/PHP81/Php81LanguageFeaturesTest.php
index 52b9933..59eadaf 100644
--- tests/PHP81/Php81LanguageFeaturesTest.php
+++ tests/PHP81/Php81LanguageFeaturesTest.php
@@ -92,6 +92,27 @@ class Php81LanguageFeaturesTest extends MockeryTestCase
         $this->expectException(RuntimeException::class);
         $mock->throws();
     }
+
+    /** @test */
+    public function it_can_mock_a_class_with_a_never_returning_type_hint_with_exit()
+    {
+        $mock = Mockery::mock(NeverReturningTypehintClass::class)->makePartial();
+
+        $pid = pcntl_fork();
+
+        if (-1 == $pid) {
+            $this->markTestSkipped("Couldn't fork for exit test");
+
+            return;
+        } else if ($pid) {
+            pcntl_waitpid($pid, $status);
+            $this->assertEquals(123, pcntl_wexitstatus($status));
+
+            return;
+        }
+
+        $mock->exits();
+    }
 }
 
 interface LoggerInterface
@@ -154,7 +175,7 @@ class NeverReturningTypehintClass
 
     public function exits(): never
     {
-        exit;
+        exit(123);
     }
 }
 class IntersectionTypeHelperClass
```

ghostwriter:
That's perfect @davedevelopment👌🏾, Thank you!

tatupesonen:
Thank you!

Issue #1190, opened 2022-09-07
Since v1.5.1 I'm getting "This test did not perform any assertions"


simPod:
I compared 1.5.1 to 1.5.0 and it seems the only function change is this https://github.com/mockery/mockery/pull/1180

I'm getting "This test did not perform any assertions" for tests that contain `->shouldReceive` calls.

It was counted towards assertions in all previouis Mockery versions.

simPod:
And my main issue is that from the changelog I'm unable to find out what should I do.

andrewbroberg:
Noticing the same issue here, could it be this change that has caused it potentially?

https://github.com/mockery/mockery/commit/72cf17bb9ac96fe52015c8fbc41b2a6b88e3177a

andrewbroberg:
Found the reason why these are failing, it looks like it may be by design.

```php
Mockery::mock(Foo::class)->shouldReceive('bar');
```

The above will fail as we have not provided the number of calls we expect to it, we will need to update it to something like below

```php
Mockery::mock(Foo::class)->shouldReceive('bar')->once();
```

The added test in the commit show this expected behaviour https://github.com/mockery/mockery/commit/72cf17bb9ac96fe52015c8fbc41b2a6b88e3177a#diff-e290c245179d001a519df9ee5cecf659df2e2ff8506829ef10dcc6900559fe56

simPod:
@andrewbroberg that's nice actually. I guess it's a fix then?

ghostwriter:
correct, `->shouldReceive(…)` would result in zero assertions because it's only stubbing the method call.

GrahamCampbell:
I think that is debatable. That should probably be regarded as a breaking change, and delayed till 2.x, and reverted in 1.x.

davedevelopment:
Definitely debatable, but nobody turned up to the [debate](https://github.com/mockery/mockery/pull/1180#issuecomment-1137008454) and then I forgot...

ghostwriter:
Regarding BC Break,

could/should we internally set the default expectation for method stubs `->shouldReceive(…)`  to `->once()` and allow it to be overwritten?

- `->shouldReceive(…)` would result in one assertion, addressing the issue that #1180 fixed.

``` php
$mock = Mockery::mock(); 
$mock->shouldReceive('foo'); // Same as: $mock->shouldReceive('foo')->once();
// can be overwritten
// $mock->shouldReceive('foo')->atLeast()->once();
[...]

PR #1192, opened 2022-09-27
Update the changelog to include changes from #1180

downsider:
In response to issue #1190, I have updated the documentation to record the changes from #1180 

I haven't labelled the change as a BC break to avoid contention; the scope of this MR is purely to make people aware that the change happened. To that end, I hope we can release this under 1.5.2 quickly, even if this is the only change.


davedevelopment:
Sorry, shouldn't have missed that one.

PR #1210, opened 2023-01-31
Add support for target class attributes

rneudorf-godaddy:
Resolves #1209.

This PR adds support for target class attributes. The `#[\AllowDynamicProperties]` attribute is the most straightfoward way to resolve PHP 8.2 compatibility issue mentioned in #1209.

### `TargetClassInterface` Update
- Added `getAttributes()` which returns an array of attributes
- Updated `DefinedTargetClass` to return an empty array.
- Updated `UndefinedTargetClass` to return `\AllowDynamicProperties` 👈 i.e. the fix

### Added `ClassAttributesPass`
- This pass inserts the attributes just before the class declaration.
- Updated `StringManipulationGenerator` to use this pass.

### Test Updates
- A few test were taking dynamically declaring properties in their mocks, updated those. 
- Added a test for `ClassAttributesPass`

PR #1214, opened 2023-02-19
Allow Dynamic Properties on PHP 8.2

ghostwriter:
If accepted, this patch allows Mockery to add `#[\AllowDynamicProperties]` PHP Attribute to the mock classes it generates.

before: 
```php
class Mockery_Foo {
    // snip
}
```
after:
```php
#[\AllowDynamicProperties]
class Mockery_Foo {
    // snip
}
```


davedevelopment:
Might be able to take care of this in #1210

Issue #962, opened 2019-03-26
Mockery 2.0 discussion


davedevelopment:
Should there be a Mockery 2.0? 

As one of the maintainers, I have zero to little interest in maintaining Mockery's more complicated and esoteric features, such as [instance mocks](http://docs.mockery.io/en/latest/reference/instance_mocking.html), [demeter chains](http://docs.mockery.io/en/latest/reference/demeter_chains.html), [class constants](http://docs.mockery.io/en/latest/cookbook/class_constants.html) etc. 

I occasionally might use [partial mocks](http://docs.mockery.io/en/latest/reference/partial_mocks.html), but that's very rare.

What I do a lot of is basic stubbing and spying, with the occasional bit of mocking. My preference for Mockery 2.0, would be to cut back on all of the aforementioned features and concentrate on making the basics really good, but that would probably be a different library altogether, as it would significantly break BC with previous versions. It wouldn't just be a case of changing syntax, Mockery would no longer support the paradigms it previously supported, meaning tests would probably need to be completely rewritten.

There's also the question of whether mockery is needed anymore. PHPUnit's [built in doubles](https://phpunit.readthedocs.io/en/8.0/test-doubles.html) are a lot better now and they also shipped first class support for [prophecy](https://phpunit.readthedocs.io/en/8.0/test-doubles.html#prophecy), so some people will turn to that too. Mockery's more esoteric features are probably one of the few things that makes it stand out from PHPUnit Doubles and Prophecy, but the better you get at testing your code, they less likely you are to use them.

mfn:
> There's also the question of whether mockery is needed anymore. PHPUnit's [built in doubles](https://phpunit.readthedocs.io/en/8.0/test-doubles.html) are a lot better now and they also shipped first class support for [prophecy](https://phpunit.readthedocs.io/en/8.0/test-doubles.html#prophecy)

There's a detail which made be prefer Mockery over the years: if you define a mock but don't add expectations, Mockery fails.

To me this is an invaluable default to me. PHPUnits built in behaviour simple returns `null` on such calls and you don't need to define expectations for calls. Leading to all sorts of hidden/missing things in your tests, even if you're super careful.

PS: never used prophecy, though

cbernard73:
I can accomplish the same thing in either Mockery or the built-in doubles, but Mockery is, in my opinion, far more elegant.  Here's an example of using Mockery to mock a service that provides a fluent interface:

```
$mService = $this->mockery->mock(MyService::class);
[...]

PR #1219, opened 2023-02-23
Support PHPUnit 10.x

ghostwriter:
Hey @davedevelopment, (and contributors)

currently working on a `Mockery Extension` for PHPUnit 10 (which explicitly requires `PHP 8.1`).

---

Mockery needs to do a few things: 

- Register a new `MockeryExceptionInterface` as a `"Failure Type"` (to be implemented by all `Mockery` Exception classes) 
> As of https://github.com/sebastianbergmann/phpunit/commit/74ce5b661a74b488e4a515256b9f35c5a8562683, all exceptions thrown in third-party subscribers are ignored.

> to communicate to `PHPUnit` the result of the `Mockery Expectation verification`; making sure that it is correctly interpreted (as a failure, not as an error) via `TestCase::registerFailureType()` implemented in https://github.com/sebastianbergmann/phpunit/commit/28766f5aac456f8093a1e946aad79c9cf96bb0ca for `PHPUnit 10.1`.

-  Exclude Mockery Directory via `PHPUnit\Util\ExcludeList::addDirectory`

- Add expectations to be verified via `TestCase::addToAssertionCount(Mockery::getContainer()->mockery_getExpectationCount());` before the test starts.

- Verify expectations after the test has finished (checks if `Mockery::close()` has been called, and let the test fail if it hasn't.)

---

How does it work: 

- Bootstrap `\Mockery\Adapter\Phpunit\Extension` class in `phpunit.xml` or `phpunit.xml.dist`.

```xml
<extensions>
    <bootstrap class="\Mockery\Adapter\Phpunit\Extension"/>
</extensions>
```

- Extend `\Mockery\Adapter\Phpunit\MockeryTestCase` class in your tests.

```php
<?php

declare(strict_types=1);

use Mockery\Adapter\Phpunit\MockeryTestCase;

final class CustomTest extends MockeryTestCase
{
    //  snip
}
```

---

Deprecated and removed features:

- `TestListener`
- `Testcase::markAsRisky()`

---

Do we have any plans to drop support for older versions of PHP?

- `PHPUnit 10` uses `PHP 8.1` features (eg. `never` return type declarations which would fail on older versions of PHP.)

Should we follow, a similar path as PHPUnit and raise the minimum required version of PHP to 8.1 for Mockery v2?

What's our desired version path for adding `Mockery Extension` for  `PHPUnit 10`:

- a new minor (v1.6)
- a new major (v2.x)

should we resurrect [Mockery 2.0 discussion](https://github.com/mockery/mockery/issues/962#top)?


davedevelopment:
> Should we follow, a similar path as PHPUnit and raise the minimum required version of PHP to 8.1 for Mockery v2?
> 
> What's our desired version path for adding Mockery Extension for PHPUnit 10:
> 
> a new minor (v1.6)
> a new major (v2.x)

I'm using Mockery exclusively with PHPUnit, so I would be happy to "follow" PHPUnit with regards to minimum PHP versions. 

As for v1.6 vs v2.x, I'm really not sure. PHPUnit integration certainly isn't a new feature, we're just trying to keep up with the jones', but the integration has changed so much over the years, it feels like a clean break would make sense for signalling. We could say v2 is for PHPUnit 10, and maybe going forward we could have v3 for PHPUnit 11 and so on. I think PHPUnit only ever support 3 versions at max, so we could follow that too. 

> should we resurrect https://github.com/mockery/mockery/issues/962#top?

I think that discussion is probably dead, I'll close it.

composer.json
```diff
@@ -34,12 +34,12 @@
         "ext-pcre": "*"
     },
     "require-dev": {
-        "phpunit/phpunit": "^8.5 || ^9.6",
+        "phpunit/phpunit": "^10.1",
```
simPod, line 37:
```suggestion
        "phpunit/phpunit": "^10.1 || ^11",
```

should we add v11 as well?

maartenderie:
@ghostwriter @simPod what is the next step for this PR? Thanks in advance

I'm trying to deal with CVE related to phpunit and mockery seems to be holding this up (and it's in a lot of places).

simPod:
Seems stale. I just did review...

Issue #1033, opened 2019-12-27
Improve exception debugging


davedevelopment:
We had something going in #1000, but reverted it due to #1026, it would be good to try and get an implementation going that everyone can get along with, if not, something for more recent versions of PHP.

GrahamCampbell:
Maybe it could go into Mockery 2.0, which could be PHP 7.0+ only?

davedevelopment:
I think that's likely :+1:

davedevelopment:
Another issue raised: #1037

Issue #936, opened 2019-01-04
Diff actually called method args on InvalidCountException


gn-ley:
I use Mockery to assert if a method inside of my test subject is called with a correct parameter. Unfortunately this parameter can be a quite complex array (hope you don’t blame me for bad design), which makes debugging tests really hard. I just get an error saying 

> Method update(<Any Arguments>) from Mockery_0_ArticleSynchronization should be called exactly 0 times but called 1 times.

I get this error because I defined that update `shouldNotReceive` a call - except the one I stated with `shouldReceive()->with()`.

What I wanted was a diff that tells me, why there was no match. What is difference between my expected args and the actual args?

After quite some research in docs and github issues, I decided I need to build my own solution. I want to share with you my solution, that you may integrate something similar into Mockery itself.

It works with Spies and its raised Exceptions, that I will evaluate with a Helper class. Mockery does track all called methods on the mocked object. Unfortunately Mockery doesn’t reveal this information to the public. This is why my Helper function needs ugly access changes with Reflections.

MockeryExceptionHelper.php
```php
<?php

namespace Helper;

class MockeryExceptionHelper
{
    /**
     * @param \Mockery\Exception\InvalidCountException $ex
     * @param null $method
     * @return mixed
     */
    public static function lastMethodCallArgs(\Mockery\Exception\InvalidCountException $ex)
    {
        $reflector = new \ReflectionObject($ex->getMock());
        $getReceivedMethodCallsMethod = $reflector->getMethod('_mockery_getReceivedMethodCalls');
        $getReceivedMethodCallsMethod->setAccessible(true);
        $receivedMethodCalls = $getReceivedMethodCallsMethod->invoke($ex->getMock());

        $reflectorInner = new \ReflectionObject($receivedMethodCalls);
        $property = $reflectorInner->getProperty('methodCalls');
        $property->setAccessible(true);
        $receivedMethodCalls = $property->getValue($receivedMethodCalls);

        /** @var $receivedMethodCalls \Mockery\MethodCall[] */
        return $receivedMethodCalls[0]->getArgs()[0];
    }
}
```

SynchronizationTest.php:
```php
class SynchronizationTest extends \Codeception\Test\Unit
{
    public function testUpdateArticle()
    {
        $sync = \Mockery::spy(ArticleSynchronization::class . "[create,update,delete]")->makePartial();
        $sync->shouldAllowMockingProtectedMethods();
        $sync->shouldNotReceive('create');
[...]

PR #1221, opened 2023-03-05
Add dump helper

tontonsb:
This PR proposes a new expectation method to help in debugging expectations. When your expectation doesn't match and you want to understand what was actually called, call `->andDump()` on your expectation like this:

```php
$connection
    ->shouldReceive('update')->andDump()
    ->with('arg1', 'arg2')
    ->once();
```

The dump method will throw an exception showing the called method, args and the stack trace.

The rest of the chain will have no effect as `andDump()` returns a clone of the expectation. Thus unlike with dumping in an `andReturn` or putting a dumping matcher inside the `with()`, this does not require commenting out or modifying your non-matching `with` and other method calls.

### DIFF
--- a/library/Mockery/Exception/Dump.php
+++ b/library/Mockery/Exception/Dump.php
@@ -1,0 +1,5 @@
+<?php
+
+namespace Mockery\Exception;
+
+class BadMethodCallException extends \Mockery\Exception {}
--- a/library/Mockery/Expectation.php
+++ b/library/Mockery/Expectation.php
@@ -616,6 +616,22 @@ class Expectation implements ExpectationInterface
     }
 
     /**
+     * Dump the args supplied by the method call.
+     *
+     * @return self
+     */
+    public function andDump()
+    {
+        $this->_closureQueue = [
+            fn(...$args) => throw new Exception(
+                'Called ' . \Mockery::formatArgs($this->_name, $args)
+            ),
+        ];
+
+        return clone $this;
+    }
+
+    /**
      * Set Exception class and arguments to that class to be thrown
      *
      * @param string|\Exception $exception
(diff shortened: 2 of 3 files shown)


davedevelopment:
I like the concept, not sure on the implementation/api though. 

As I understand, it will only work if the `andDump` is called before the arguments specification etc? 

``` php

$connection
    ->shouldReceive('update')
    ->with('arg1', 'arg2')
    ->once()->andDump(); // probably wouldn't do anything if the expectation was not matched?
```

There's no doubt this is something Mockery needs, a clear way of determining why your expectation wasn't met and has come up in the past, with things like #1033 and #936.

Perhaps, this could be a something to start with and something we ship quickly, with the later view of something much better. 

Could we think on the naming a little? I'm not against `andDump`, but wonder if we could be more specific. When I first read `andDump`, I assumed it would dump the arguments to output, but then carry on anyway, so I might see all of the calls within the test.  I can't really think of anything better to name it, without it getting lengthy, but here are a couple of ideas in case anyone else can think of something:

``` php
$connection->shouldReceive('update')->thenDumpAndThrow();
$connection->shouldReceive('update')->thenThrowDebugException();
```

tontonsb:
> As I understand, it will only work if the andDump is called before the arguments specification etc?

Yes, it's somewhat of a "let's see how the chain behaves this far". You might also do `->with(\Mockery::on($matcherCallback))->andDump()` to see the call that has been matched.

> When I first read andDump, I assumed it would dump the arguments to output, but then carry on anyway

I actually tried doing things like that, but all of my attempts seemed to affect too many parts and seeing that #1000 was reverted I decided to go with the simplest approach so it wouldn't break anything.

> in case anyone else can think of something

Yeah, a short and descriptive name would be nice :)

PR #1232, opened 2023-04-15
Split PHPUnit TestSuites for PHP 8.2

ghostwriter:
This pull request updates the `phpunit.xml.dist` configuration file to exclude tests based on the PHP version that is being used.

This will allow us to run tests for different PHP versions in isolation, which will make it easier to identify and fix issues that are specific to certain PHP versions.


ghostwriter (commented):
This pull request only updates the `phpunit.xml.dist` file and all CI checks specifically related to this PR are passing. https://github.com/mockery/mockery/pull/1232/files

Therefore, I am going to proceed and merge this pull request.

PR #1237, opened 2023-04-17
Unignore `composer.lock` file

ghostwriter:
This patch removes `composer.lock` from the `.gitignore` and adds a new lock file.


ghostwriter:
This pull request adds `composer.lock` file. https://github.com/mockery/mockery/pull/1237/files

PR #1239, opened 2023-04-18
Prevent CI run duplication

ghostwriter:
This patch limits the CI actions to only run on `push` events, on branches matching the regex to prevent duplicate runs in pull requests.


ghostwriter:
This pull request only updates the CI configuration. https://github.com/mockery/mockery/pull/1239/files

PR #1245, opened 2023-04-30
Deprecate `hamcrest/hamcrest-php` package

ghostwriter:
This patch is part of #1227, has the following changes:

 - Adds `@trigger_error('Hamcrest package has been deprecated and will be removed in 2.0', E_USER_DEPRECATED);`
 - Remove the recommendation to use Hamcrest


ruudk:
I just noticed this deprecation. Is there a Mockery equivalent for using Hamcrest's `IsEqual::equalsTo`?

ghostwriter:
Related: #1275 and #1276

PR #1253, opened 2023-05-02
Prevent array to string conversion in Subset matcher

### LINKED ISSUE #1252: Bug: nested arrays cause array to string conversion in Subset
If an expectation fails a count check, Mockery converts argument matchers into strings in order to display the expectation in the assertion failure message. When this is done to the `Subset` matcher when using a nested subset of arrays, casting to a string causes an "Array to string conversion" error

```
public function testSubsetError()
{
  $expected = [
    "foo" => [
      "bar" => "baz"
    ]
  ];

  $mock = \Mockery::mock(\Exception::class);
  $mock->shouldReceive("getMessage")->with(\Mockery::subset($expected))->once();
}
```

Results in :
```
There was 1 error:

1) tests\ExampleTest::testSubsetError()
Array to string conversion

/project/path/mockery/library/Mockery/Matcher/Subset.php:87

```

downsider:
Fixes #1252


ghostwriter (approved):
Looks good to me, going to proceed and merge this pull request.

Thank you.

PR #1275, opened 2023-06-06
Add `IsSame` Argument Matcher

ghostwriter:
This pr adds an `IsSame` [Argument Matcher](http://docs.mockery.io/en/latest/reference/argument_validation.html#additional-argument-matchers)


ghostwriter:
Psalm failure is due to legacy code, not related to this PR.

PR #1276, opened 2023-06-06
Add `IsEqual` Argument Matcher

### LINKED ISSUE #1268: What is the suggested replacement for Hamcrest `IsEqual::equalTo`
I noticed the following deprecation in our project:
> Hamcrest package has been deprecated and will be removed in 2.0

Turns out we've been using `IsEqual::equalTo` in a few (100) places. 

What's the suggested solution in these cases?

## Example

```php
$repository->expects('save')->with(IsEqual::equalTo(new User(1));
```

ghostwriter:
This pr adds an `IsEqual` [Argument Matcher](http://docs.mockery.io/en/latest/reference/argument_validation.html#additional-argument-matchers)

Resolves #1268

### DIFF
--- a/library/Mockery/Matcher/IsEqual.php
+++ b/library/Mockery/Matcher/IsEqual.php
@@ -1,0 +1,27 @@
+<?php
+
+namespace Mockery\Matcher;
+
+class IsEqual extends MatcherAbstract
+{
+    /**
+     * Check if the actual value matches the expected.
+     *
+     * @param mixed $actual
+     * @return bool
+     */
+    public function match(&$actual)
+    {
+        return $this->_expected == $actual;
+    }
+
+    /**
+     * Return a string representation of this Matcher
+     *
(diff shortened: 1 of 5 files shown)


ruudk:
Thanks @ghostwriter 💙 

Would it be possible to tag this?

oprypkhantc:
@ghostwriter Would a PR adding a factory method for new `IsEqual` and `IsSame` matchers to `Mockery` be welcomed, similar to how `Mockery::any()` exist? i.e.
 
```php
class Mockery {
    public static function equal($other) {
        return new IsEqual($other);
    }
}
```

ghostwriter:
Hey @oprypkhantc,

As you requested, I have included the factory methods for both `IsEqual` and `IsSame` in pull request #1336.

By the way, for future reference, please create a new issue so that we can keep old and new issues and requests separate.

Thanks

PR #1294, opened 2023-07-13
Correct return types of Mock for phpstan

antonsacred:
This PR should fix old problem when phpstan stop working correctly with Mockery when you do something like `$mock = Mockery::mock(SomeClass::class)->shouldIgnoreMissing();`

### DIFF
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -1914,8 +1914,8 @@
       <code>shouldNotReceive</code>
     </PossiblyUnusedMethod>
     <PossiblyUnusedReturnValue>
-      <code>Mock</code>
       <code>\Mockery\ExpectationDirector|null</code>
+      <code>static</code>
     </PossiblyUnusedReturnValue>
     <UndefinedDocblockClass>
       <code>null|array|Closure</code>
--- a/src/Mockery/LegacyMockInterface.php
+++ b/src/Mockery/LegacyMockInterface.php
@@ -47,12 +47,12 @@ interface LegacyMockInterface
     /**
      * Set mock to ignore unexpected methods and return Undefined class
      * @param mixed $returnValue the default return value for calls to missing functions on this mock
-     * @return $this
+     * @return static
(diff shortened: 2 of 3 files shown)


ghostwriter:
Hey @antonsacred,

Thank you for your contribution. 

I appreciate the time you invested in preparing this pull request.

However, I have decided not to merge it because adding the `@return $this` doc comment is not the correct way to address this issue.

There is a [PHPStan extension for Mockery](https://github.com/phpstan/phpstan-mockery), have you tried using it?

If you still have the same issue, with the extension enabled, then we can update the doc comment to `@return MockInterface` instead.

What do think?

antonsacred:
Thank you for the answer. Here is short code which shows the problem with phpstan. I use latest version of phpstan and extension.

```php
function test(\stdClass $object): \stdClass
{
        return $object;
}

$mock = m::mock(\stdClass::class)->shouldIgnoreMissing();

$result = $this->test($mock);
```

This code example will have phpstan error `Parameter #1 $object of method test() expects stdClass, Mockery\MockInterface given.`

However if we change return type to `$this` - phpstan will understand that mock object wasn't mutated

ghostwriter:
> What do you prefer more `@return static` or `@return $this` ?

Let’s go with `@return static` since most tools and IDEs support this syntax.

To fix the currently failing psalm issue, run `composer psalm:baseline` and commit any changes in `psalm-baseline.xml`.

antonsacred:
Changed `$this` to `static`.
Committed changes in psalm-baseline.xml which are related to this PR.

ghostwriter (approved):
Looks good to me, going to proceed and merge this pull request.

Thank you.

ghostwriter:
you may use the branch alias `1.6.x` until it is released. 

`composer require mockery/mockery:1.6.x --dev`

ghostwriter:
Released https://github.com/mockery/mockery/releases/tag/1.6.3

PR #1299, opened 2023-07-16
Test PHP `8.3` language features

### LINKED ISSUE #1291: Support PHP `8.3`
This issue is for tracking PHP 8.3 feature that we may need support and [test](https://github.com/mockery/mockery/blob/c322a33e9461d0c02514b815593131d825bb1de3/.github/workflows/tests.yml#L18-L20) for compatibility.

- [x] [`Typed Class Constants`](https://wiki.php.net/rfc/typed_class_constants) can be declared in classes, interfaces, traits, and enums.

```php
enum Enums {
    const string FOO = "bar";
}

class Classes implements I { 
    const string BAR = E::FOO;
}

interface Interfaces {
    const string BAR = E::FOO;
}
 
trait Traits {
    const string BAR = E::FOO;
}
```

- Class constant type declarations support all PHP type declarations except `void`, `callable`, and `never`.
- Class constants are covariant, meaning that their types are not allowed to widen during inheritance.
- Constant values must match the type of the class constant, with the exception that `float` class constants can also accept `integer` values.
- [ReflectionClassConstant](https://www.php.net/manual/en/class.reflectionclassconstant.php) is extended with two methods: getType() and hasType().

```php
class Foo {
    public const string BAR = 'baz';
}
 
class Bar extends Foo {
    // Doesn't work anymore! we cannot change the type and assign a value with a different type.
    public const array BAR = ['foo', 'bar', 'baz'];
    // OK.
    public const string BAR = 'foo';
}
```

- [x] [`Dynamic class constant fetch`](https://wiki.php.net/rfc/dynamic_class_constant_fetch) allow class constants to be accessed dynamically using variables.

Instead of accessing class constants with a static string value (e.g. ClassName::CONSTANT), we can use a variable containing the constant name.

```php
class ClassName {
    public const CONSTANT = 42;
}

$constant = 'CONSTANT';
 
ClassName::{$constant}
```

---

> **Note**
> Milestone TBD by Jul 18, 2023 (PHP `8.3` Feature freeze)

Resources:

- https://wiki.php.net/rfc#php_83
- https://wiki.php.net/todo/php83

ghostwriter:
Resolves #1291

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -69,6 +69,7 @@
     "autoload-dev": {
         "psr-4": {
             "Fixture\\": "tests/Fixture/",
+            "Mockery\\Tests\\Unit\\": "tests/Unit",
             "test\\": "tests/"
         },
         "files": [
--- a/composer.lock
+++ b/composer.lock
@@ -1279,16 +1279,16 @@
         },
         {
             "name": "phpstan/phpdoc-parser",
-            "version": "1.22.1",
+            "version": "1.23.0",
             "source": {
                 "type": "git",
                 "url": "https://github.com/phpstan/phpdoc-parser.git",
-                "reference": "65c39594fbd8c67abfc68bb323f86447bab79cc0"
+                "reference": "a2b24135c35852b348894320d47b3902a94bc494"
             },
             "dist": {
                 "type": "zip",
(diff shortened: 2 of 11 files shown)


ghostwriter:
Run `make test-83` to test.

PR #1301, opened 2023-07-17
Fix mocking classes with `new` initializers in method and attribute params on PHP 8.1

### LINKED ISSUE #1300: Mock classes with new in methods on PHP 8.1
### Mockery Version

1.6.2

### PHP Version

PHP 8.2

### Issue Description

Hey,

I'm trying to mock a class method which has a `new DateTime` as default argument but it's not working. I get: `Error: Undefined constant "DateTime"`.

### Steps to Reproduce

```php
<?php

declare(strict_types=1);

namespace Tests;

use DateTime;
use Mockery;
use PHPUnit\Framework\TestCase;

class HandlerClassTest extends TestCase
{
    public function testItDoesNotWork(): void
    {
        $class = Mockery::mock(MockClass::class)->shouldReceive('test')->with('test')->andReturn('yay')->getMock();

        self::assertSame('yay', (new HandlerClass())->doStuff($class));
    }
}

class MockClass
{
    public function test(string $a, DateTime $b = new DateTime()): string
    {
        return $a . ' - ' . $b->format('Y-m-d');
    }
}

class HandlerClass
{
    public function doStuff(MockClass $mockClass): string
    {
        return $mockClass->test('test');
    }
}
```

### Expected Behavior

_No response_

### Actual Behavior

_No response_

### Exception or Error

```shell
docker compose exec php vendor/bin/phpunit tests/HandlerClassTest.php

PHPUnit 10.1.3 by Sebastian Bergmann and contributors.

Runtime:       PHP 8.2.7
Configuration: /var/www/html/phpunit.xml

E                                                                   1 / 1 (100%)

Time: 00:00.006, Memory: 12.00 MB

There was 1 error:

1) Tests\HandlerClassTest::testItDoesNotWork
Error: Undefined constant "DateTime"

/var/www/html/tests/HandlerClassTest.php:33
/var/www/html/tests/HandlerClassTest.php:17

--

There was 1 risky test:

1) Tests\HandlerClassTest::testItDoesNotWork
This test did not perform any assertions

/var/www/html/tests/HandlerClassTest.php:13
```

### Additional Information

I found this PR https://github.com/mockery/mockery/pull/1160 but I think this did only work by accident? Or am I missing something? :smile:

ghostwriter:
Resolves #1300

### DIFF
--- a/src/Mockery/Generator/StringManipulation/Pass/MethodDefinitionPass.php
+++ b/src/Mockery/Generator/StringManipulation/Pass/MethodDefinitionPass.php
@@ -73,9 +73,7 @@ class MethodDefinitionPass implements Pass
                         if ($isPhp81) {
                             if (enum_exists($prefix)) {
                                 $prefix = var_export($defaultValue, true);
-                            }
-
-                            else if (
+                            } elseif (
                                 !$param->isDefaultValueConstant() &&
                                 // "Parameter #1 [ <optional> F\Q\CN $a = new \F\Q\CN(param1, param2: 2) ]
                                 preg_match(
--- a/src/Mockery/Generator/StringManipulation/Pass/MethodDefinitionPass.php
+++ b/src/Mockery/Generator/StringManipulation/Pass/MethodDefinitionPass.php
@@ -77,7 +77,7 @@ class MethodDefinitionPass implements Pass
                                 !$param->isDefaultValueConstant() &&
                                 // "Parameter #1 [ <optional> F\Q\CN $a = new \F\Q\CN(param1, param2: 2) ]
                                 preg_match(
-                                    '/<optional>\s.*?\s=\snew\s(.*?)\s]/',
+                                    '#<optional>\s.*?\s=\snew\s(.*?)\s]$#',
                                     $param->__toString(),
                                     $matches
                                 ) === 1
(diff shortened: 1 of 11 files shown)


src/Mockery/Generator/StringManipulation/Pass/MethodDefinitionPass.php
```diff
@@ -70,8 +70,20 @@ protected function renderParams(Method $method, $config)
...
-                            $prefix = var_export($defaultValue, true);
+                        if ($isPhp81) {
+                            if (enum_exists($prefix)) {
+                                $prefix = var_export($defaultValue, true);
+                            } elseif (
+                                !$param->isDefaultValueConstant() &&
+                                // "Parameter #1 [ <optional> F\Q\CN $a = new \F\Q\CN(param1, param2: 2) ]
+                                preg_match(
+                                    '#<optional>\s.*?\s=\snew\s(.*?)\s]$#',
+                                    $param->__toString(),
```
Bilge, line 81: It is not intended to ever call `__toString()` directly. This should be a cast instead (`(string)$param`).

ghostwriter: @Bilge I agree,

However the reason for using `__toString()` instead was because `$param` is actually an object that is proxying these calls via `__call` method to the `ReflectionParameter` which was giving an error because the `__toString` expects a string return type.

https://github.com/mockery/mockery/blob/b1be135c1ba7632f0248e07ee5e6e412576a309d/src/Mockery/Generator/Parameter.php#L28

PR #1312, opened 2023-07-26
Mock disjunctive normal form types

### LINKED ISSUE #1310: Can't accurately mock Disjunctive Normal Form Type
# Mockery Version

1.6.4

# PHP Version

PHP 8.2

# Issue Description

## Summary
From PHP 8.2, a type called 'Disjunctive Normal Form' has been introduced. This is designed to represent complex types that connect ANDs with ORs.

When Mockery analyzes the class to be mocked and outputs the methods to the mock object, it doesn't take this Disjunctive Normal Form into account, and the type of arguments changes between the actual implementation and the mock object's implementation.

For example, when mocking a method foo() that takes a type of A|(B&C), the arguments of foo() implemented in the actual mock object become A|B|C.

RFC link: https://wiki.php.net/rfc/dnf_types

## Relevant part of the implementation
Looking at the implementation of typeToString(), it determines whether to join with & or | depending on whether the type of the argument is IntersectionType or not. However, since getTypeInformation() initially flattens nested types into a one-dimensional array (actually, a list of associative arrays with keys typeHint and isPrimitive), when they are joined back together in typeToString(), the nesting is resolved and a type like A|(B&C) becomes A|B|C.

https://github.com/mockery/mockery/blob/f14c1b124e7c8901bf70dc6eafa290179dfac319/library/Mockery/Reflector.php#L122-L128

# Steps to Reproduce

Execute this code

```php
use Mockery;

class A {};
class B {};
class C {};

class Sample
{
	public function foo(A|(B&C) $arg) {
		var_dump($arg);
	}
}

$mock = Mockery::mock(Sample::class);

$rc = new \ReflectionClass($mock);

$type = $rc->getMethod('foo')
	->getParameters()[0]
	->getType();

var_dump($type->__toString());
```

# Expected Behavior

output like this

```
string(5) "A|(B&C)"
```

# Actual Behavior

output like this

```
string(5) "A|B|C"
```

# Exception or Error

_No response_

# Additional Information

Thank you all for your continuous development of Mockery. I came across a bug while reading the source code for a different topic, and I thought I would report it. I considered fixing it myself and submitting a PullRequest, but it looks like it might take some time, so I'm just reporting the issue for now.

ghostwriter:
Resolves #1310


ghostwriter (commented):
and Add the rest of the examples from https://wiki.php.net/rfc/dnf_types in the test fixtures.

library/Mockery/Reflector.php
```diff
@@ -205,14 +189,78 @@ private static function getTypeInformation(\ReflectionType $type, \ReflectionCla
...
+                $type->getTypes()
+            );
+
+            return implode(
+                '&',
+                array_map(
+                    fn (string $type): string => $type,
+                    $types
+                )
+            );
```
fuwasegu, line 246:
```suggestion
            return implode(
                '&',
                $types,
            );
```

II was going to suggest this, but maybe you wrapped it in array_map() because you want to guarantee in Runtime that $types is a list<string>?

Personally, I think @var list<string> $types type hinting would be fine.

PR #1313, opened 2023-07-29
Type declaration for `iterable|object`

### LINKED ISSUE #1311: Interpreting iterable|object as array|Traversable|object results in a runtime error.
# Mockery Version

1.6.4

# PHP Version

PHP 8.2

# Issue Description

## Introduction
This bug, as can be seen in the following php-src issue, is not solely a problem with Mockery but involves a bug in PHP itself. Therefore, if this project has made the decision to adhere to the PHP implementation, it may be acceptable to exclude this issue from being treated as a bug.

https://github.com/php/php-src/issues/11797

## Summary
In Mockery, the ReflectionClass is used to analyze the class to be mocked. When copying the methods of the class to be mocked to the Mock class, the type of the method's arguments are analyzed and outputted using ReflectionProperty. However, the output result of iterable changes depending on whether the version of PHP in the execution environment is 8.2 or not.

PHP 8.1 and below ... iterable
PHP 8.2 ... Traversable|array
This is because iterable, which was a built-in pseudo-type, has become a "type alias" that is converted to array|Traversable at compile time from PHP 8.2.

Reference: https://www.php.net/manual/en/language.types.iterable.php

While there is no problem with this fact itself, when Mockery analyzes the method using Reflection for a class method implemented with the argument type as iterable|object, the type of the method implemented on the mock class side becomes Traversable|object|array. This results in the following error because Traversable|object is redundant.

```
Type Traversable|object|array contains both object and a class type, which is redundant
```

Therefore, you cannot run the test with a Fatal Error when eval() in Mockery's EvalLoader.

If this bug is to be addressed in Mockery, I believe that when the argument type contains array|Traversable, it will need to be replaced with iterable or similar corrections."

# Steps to Reproduce

_No response_

# Expected Behavior

_No response_

# Actual Behavior

_No response_

# Exception or Error

_No response_

### Additional Information

_No response_

ghostwriter:
Resolves #1311

### DIFF
--- a/library/Mockery/Reflector.php
+++ b/library/Mockery/Reflector.php
@@ -15,8 +15,8 @@ namespace Mockery;
  */
 class Reflector
 {
-    private const TRAVERSABLE_OBJECT_ARRAY = ['\Traversable','object', 'array'];
-    private const ITERABLE_OBJECT = ['iterable','object'];
+    private const TRAVERSABLE_ARRAY = ['\Traversable', 'array'];
+    private const ITERABLE = ['iterable'];
     /**
      * Determine if the parameter is typed as an array.
      *
@@ -251,9 +251,9 @@ class Reflector
                 $type->getTypes()
             );
 
-            $intersect = array_intersect(self::TRAVERSABLE_OBJECT_ARRAY, $types);
-            if (self::TRAVERSABLE_OBJECT_ARRAY === $intersect) {
-                $types = self::ITERABLE_OBJECT + array_diff($types, self::TRAVERSABLE_OBJECT_ARRAY);
+            $intersect = array_intersect(self::TRAVERSABLE_ARRAY, $types);
+            if (self::TRAVERSABLE_ARRAY === $intersect) {
+                $types = self::ITERABLE + array_diff($types, self::TRAVERSABLE_ARRAY);
             }
 
             return implode(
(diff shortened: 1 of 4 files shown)


library/Mockery/Reflector.php
```diff
@@ -249,6 +251,11 @@ private static function getTypeFromReflectionType(\ReflectionType $type, \Reflec
                 $type->getTypes()
             );
 
+            $intersect = array_intersect(self::TRAVERSABLE_OBJECT_ARRAY, $types);
+            if (self::TRAVERSABLE_OBJECT_ARRAY === $intersect) {
```
ghostwriter, line 255: I'm trying to determine a more appropriate question to ask in this context. 

Instead of asking if the list of `$types` contains all the values of `TRAVERSABLE_OBJECT_ARRAY`, and if so, replacing it with the value of `ITERABLE_OBJECT` and merging any other values from `$types`

should we ask if the list of `$types` contains the value `object` and any type that is prefixed with `\`, indicating it is a class(object)?

ghostwriter: > I think the "easiest" way to fix that one, is just to keep track if both `\Traversable` and `array` are in the type list and just drop those and convert them to `iterable`.
>
> But yeah you could also ask that question, that if you encounter `object` you can/must discard other class-types
>
_Originally posted by @girgias in [discord](https://discord.com/channels/356354025865740288/862798982936789013/1134848809885765683)._

PR #1329, opened 2023-08-14
Prevent `shouldNotReceive` from getting overridden by invocation count methods

### LINKED ISSUE #1328: shouldNotReceive is overridden by invocation count methods
### Mockery Version

1.6.4

### PHP Version

PHP 8.2

### Issue Description

When using `shouldNotReceive`, it is possible to also use `once()`, `twice()` and other invocation counting methods. These methods override the previously set invocation count, making the expectation highly unintuitive

Consider the following examples:
```php
$mock->shouldNotReceive("foo")->once(); // tests method was called once
$mock->shouldReceive("foo")->once()->twice(); // tests method was called twice
$mock->shouldNotReceive("foo")->once()->twice()->times(3); // not once, not twice but three times
```

I would say that these expectations do not make semantic sense; indeed, using `shouldNotReceive()` with an invocation counter could give the false impression that the count is negated, e.g. in the first example, that the method should not be called once, but twice or more is fine.

### Steps to Reproduce

```php
$mock = \Mockery::mock(DateTime::class);
$mock->shouldNotReceive("format")->once();

$mock->format("Y");
```

### Expected Behavior

Test should fail with an invocation count error

### Actual Behavior

Test passes

### Exception or Error

_No response_

### Additional Information

_No response_

ghostwriter:
Resolves #1328 

Currently, to indicate that an expectation is `never` expected to be called, Mockery internally sets the invocation count to `0` using `Mocker\Expectation::never()` which is equivalent to `times(0)`.

So adding additional `once()`,`twice()`, `times(n)` invocation count methods will change the expectation behavior from `shouldNotReceive` to `shouldReceive`.

a quick solution was to throw an Expectation.

```php
throw new \InvalidArgumentException(
'shouldNotReceive() method does not accept chaining additional invocation count methods.'
);
```

Should we handle this differently?

### DIFF
--- a/library/Mockery/Expectation.php
+++ b/library/Mockery/Expectation.php
@@ -41,6 +41,13 @@ class Expectation implements ExpectationInterface
     protected $_because = null;
 
     /**
+     * Expected count of calls to this expectation
+     *
+     * @var int
+     */
+    protected $_expectedCount = -1;
+
+    /**
      * Arguments expected by this expectation
      *
      * @var array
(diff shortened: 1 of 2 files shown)


ghostwriter:
I left this pull request open but received no response for discussion, again.

I will merge it now, but welcome any feedback post-merge.

Thanks!

Danny-Smart:
No news is good news :smile:

I found no fault when I checked this before

PR #1338, opened 2023-11-11
Support PHPUnit constraints as matchers

### LINKED ISSUE #1337: support for PHPUnit constraints
Now that Hamcrest support is not official, would a PR adding support for PHPUnit constraints be welcomed? Sure, Mockery has some built-in matchers, but PHPUnit has a lot more to offer: https://github.com/sebastianbergmann/phpunit/tree/main/src/Framework/Constraint

_Originally posted by @oprypkhantc in https://github.com/mockery/mockery/issues/1006#issuecomment-1806840998_

oprypkhantc:
Closes #1337. PHPUnit constraints are automatically wrapped in a constraint matcher when you pass them into an expectation, similar to how Hamcrest integration used to work.

Feel free to edit it as you wish.


oprypkhantc:
Also, can we deprecate `MatcherAbstract` in favor of `Matcher`? I had to keep `MatcherAbstract` to preserve BC, but there's really no reason to keep it as it does very little, especially now with promoted properties in PHP 8.

library/Mockery/Matcher/MatcherInterface.php
```diff
@@ -10,14 +12,16 @@
...
-interface Matcher
+interface MatcherInterface
 {
     /**
      * Check if the actual value matches the expected.
      * Actual passed by reference to preserve reference trail (where applicable)
      * back to the original method parameter.
      *
-     * @param mixed $actual
+     * @template TActual
```
oprypkhantc, line 22: This template type is not useful in any way. It's the same as just mixed @ghostwriter

ghostwriter: This template type is to used by static analysis tools like psalm, to infer the exact `type` of the argument that is being passed in.

oprypkhantc: Yes I'm aware :) But in this case you've only used it for one of the parameters, meaning it's actual type is inferred on every call but isn't used anywhere, so it's as good as `@param mixed $actual` in this case, no different.

ghostwriter:
Great work on this @oprypkhantc, Thank you.


I only made a few minor changes.

- we should support/fix the expectation messages. (Maybe in a separate PR)

- I'll figure out the issue causing the ci failure that is unrelated to this pr. (tests seem to pass when run in isolation.)

- we'll wait to hear some feedback from the community about deprecating the `MatcherAbstract` (3 business days)

library/Mockery/Matcher/MatcherInterface.php
ghostwriter: You're right, it only used in this case, dropped the template tag.

thanks for your feedback.

library/Mockery/Matcher/MatcherAbstract.php
```diff
@@ -10,7 +10,10 @@
 
 namespace Mockery\Matcher;
 
-abstract class MatcherAbstract
+/**
+ * @deprecated Implement \Mockery\Matcher\MatcherInterface instead of extending this class
+ */
```
ghostwriter, line 15:
```suggestion
 * @see https://github.com/mockery/mockery/pull/1338
 */
```

PR #1342, opened 2023-11-24
Add `.readthedocs.yml` configuration

ghostwriter:
This patch resolves failing documentation build process for docs.mockery.io, via the following changes:

 - [Add .readthedocs.yml](https://github.com/mockery/mockery/commit/f1f12c42abcd559e0af2d1413c097a5094a477fd)
 - [Add requirements.txt](https://github.com/mockery/mockery/commit/9926379aee4d564a666c527b81a9e80122387725)
 - [Update conf.py](https://github.com/mockery/mockery/commit/4260ba611b58b7ae8a01b394b06f02f75c6d00d3)

image

PR #1343, opened 2023-11-24
Update documentation default theme and build version

ghostwriter:
Follows #1342


Bilge:
Thank God for this. That default theme was awful, but this was a regression anyway. It always _used_ to look like this "new" theme.

PR #1354, opened 2024-02-05
Add array-like objects support to hasKey/hasValue matchers

### LINKED ISSUE #1353: [Feature request] Add array-like objects support to argument matchers
Hello,

I'd like to propose to add an array-like objects support to some argument matchers. 

**Motivation**
Sometimes, it is more convenient to work with array-like objects than arrays (for example, Configuration object pattern). PHP provides different possibilities to implement them, including ArrayAccess interface and built-in ArrayObject class (which implements the interface). 

**Proposal**
Add the possibility to treat array-like objects as arrays in some argument matchers (`hasKey`, `hasValue`, `subset`).

kudashevs:
Hello,

This is a draft version for #1353 that adds support of array-like objects to hasKey/hasValue matchers.

How it works. We use a guard clause to filter out values that are not arrays and objects (probably, the constraints should be more severe and besides arrays, it should accept the `ArrayAccess` implementations only). Then, we just cast the object to an array and perform the matching. For right now, it works with objects that implement the `ArrayAccess` interface and objects with public properties (for example, it can process an `stdClass` instance with some filled properties). 

If the feature looks interesting, I can elaborate it a little bit (including subset matcher).

### DIFF
--- a/library/Mockery/Matcher/HasKey.php
+++ b/library/Mockery/Matcher/HasKey.php
@@ -24,7 +24,7 @@ class HasKey extends MatcherAbstract
             return false;
         }
 
-        return array_key_exists($this->_expected, $actual);
+        return array_key_exists($this->_expected, (array)$actual);
     }
 
     /**
--- a/library/Mockery/Matcher/HasKey.php
+++ b/library/Mockery/Matcher/HasKey.php
@@ -20,7 +20,7 @@ class HasKey extends MatcherAbstract
      */
     public function match(&$actual)
     {
-        if (!is_array($actual) && !is_object($actual)) {
+        if (!is_array($actual) && !$actual instanceof \ArrayAccess) {
             return false;
         }
 
--- a/library/Mockery/Matcher/HasKey.php
+++ b/library/Mockery/Matcher/HasKey.php
@@ -20,6 +20,10 @@ class HasKey extends MatcherAbstract
      */
     public function match(&$actual)
     {
+        if (!is_array($actual) && !is_object($actual)) {
+            return false;
+        }
+
         return array_key_exists($this->_expected, $actual);
     }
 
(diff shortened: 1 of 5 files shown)


library/Mockery/Matcher/HasKey.php
```diff
@@ -20,7 +20,11 @@ class HasKey extends MatcherAbstract
      */
     public function match(&$actual)
     {
-        return array_key_exists($this->_expected, $actual);
+        if (!is_array($actual) && !is_object($actual)) {
```
ghostwriter:
```suggestion
        if (!is_array($actual) && !$actual instanceof \ArrayAccess) {
```

library/Mockery/Matcher/HasValue.php
```diff
@@ -20,7 +20,11 @@ class HasValue extends MatcherAbstract
      */
     public function match(&$actual)
     {
-        return in_array($this->_expected, $actual);
+        if (!is_array($actual) && !is_object($actual)) {
```
ghostwriter:
```suggestion
        if (!is_array($actual) && !$actual instanceof \ArrayAccess) {
```

ghostwriter:
Hey @kudashevs,

Thank you for your contribution! 

I will review your pull request and provide feedback or merge it if appropriate.

ghostwriter (commented):
This looks great, thank you!

Before merging this pull request;

1. Let’s make sure the `$actual` object implements `ArrayAccess` by using:

`!$actual instanceof \ArrayAccess`

2. Run the command `composer psalm:baseline` and commit the `psalm-baseline.xml` file.

—

Regarding objects with public properties, i’m not sure we should support that here.

It should probably be a separate matcher.  eg. `ClassHasProperty `, `ClassPropertyValue`

kudashevs:
@ghostwriter Hello again,

> 2\. composer psalm:baseline
1. Done
2. Done

Regarding objects with public properties, this is a bad practice to me (it breaks encapsulation).
It was just a consequence of a broad type and I wanted to mention it just in case. So, I totally agree.

ghostwriter:
Thank you.

These changes will be on the `1.6.x` branch until we tag `1.6.8`.

`composer require mockery/mockery:1.6.x-dev —dev`

PR #1355, opened 2024-02-05
Fix the namespace in the SubsetTest class

kudashevs:
Hello,

This PR fixes the namespace in the SubsetTest test class.


ghostwriter:
Thanks again for your contribution.

PR #1357, opened 2024-02-07
Fix 8.3 get_parent_class without arguments deprecation

kudashevs:
Hello,

This PR fixes fail of tests with `Calling get_parent_class() without arguments is deprecated` on PHP 8.3.
Just in case, here is the link to a [php.watch article](https://php.watch/versions/8.3/get_class-get_parent_class-parameterless-deprecated) with more information about this deprecation.


ghostwriter (commented):
Looks good, Thank you!

However this class will be removed in `2.0.x` (part of the clean up).

It was used to support a deprecated method between PHPUnit 8 and 9.

PR #1375, opened 2024-02-29
Refactor `library/Mockery/MockInterface.php `

ghostwriter:
This patch improves performance and readability without changing functionality.

- `library/Mockery/MockInterface.php `

### DIFF
--- a/library/Mockery/MockInterface.php
+++ b/library/Mockery/MockInterface.php
@@ -18,7 +18,7 @@ interface MockInterface extends LegacyMockInterface
      *
      * @return Expectation|ExpectationInterface|HigherOrderMessage|self
      */
-    public function allows($something = null);
+    public function allows($something = []);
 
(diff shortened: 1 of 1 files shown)


library/Mockery/MockInterface.php
```diff
@@ -5,24 +5,25 @@
...
 {
     /**
-     * @param mixed $something  String method name or map of method => return
-     * @return self|\Mockery\ExpectationInterface|\Mockery\Expectation|\Mockery\HigherOrderMessage
+     * @param mixed $something String method name or map of method => return
+     *
+     * @return Expectation|ExpectationInterface|HigherOrderMessage|self
      */
-    public function allows($something = []);
+    public function allows($something = null);
```
ghostwriter: revert 
```suggestion
    public function allows($something = []);
```

PR #1386, opened 2024-03-01
Add PHPUnit 10+ attributes

### LINKED ISSUE #1359: Deprecations about drop support Metadata doc-comment in PHPUnit 12
### Mockery Version

1.6.7

### PHP Version

PHP 8.3

### Issue Description

When run tests on latest PHPUnit 11.0.3 with latest mockery version the results return deprecation messages about using annotations and this support will be drop in PHPUnit 12

### Steps to Reproduce

Run tests on latest stable PHPUnit will return on each tests when is mockery used
```
1) Metadata found in doc-comment for method App\Tests\SomeTest::startMockery(). Metadata in doc-comments is deprecated and will no longer be supported in PHPUnit 12. Update your test code to use attributes instead.

2) Metadata found in doc-comment for method App\Tests\SomeTest::purgeMockeryContainer(). Metadata in doc-comments is deprecated and will no longer be supported in PHPUnit 12. Update your test code to use attributes instead.
```

### Expected Behavior

no deprecations

### Actual Behavior

_No response_

### Exception or Error

_No response_

### Additional Information

Is all about `@before` and `@after` annotations.. for example this on line 62 in MockeryPHPUnitIntegration.php
```
    /**
     * @before
     */
    protected function startMockery()
    {
        $this->mockeryOpen = true;
    }

    /**
     * @after
     */
    protected function purgeMockeryContainer()
    {
        if ($this->mockeryOpen) {
            // post conditions wasn't called, so test probably failed
            Mockery::close();
        }
    }
```

ruudk:
Fixes #1359

While upgrading to PHPUnit 11 I noticed the following deprecation:

Metadata in doc-comments is deprecated and will no longer be supported in PHPUnit 12. Update your test code to use attributes instead.

If we add the attribute too, it stops complaining. So this most be a way to support PHPUnit 11 and older versions.

If we ship this, we reduce a few errors. Most of the things just work fine.

### DIFF
--- a/library/Mockery/Adapter/Phpunit/MockeryPHPUnitIntegration.php
+++ b/library/Mockery/Adapter/Phpunit/MockeryPHPUnitIntegration.php
@@ -11,6 +11,8 @@
 namespace Mockery\Adapter\Phpunit;
 
 use Mockery;
+use PHPUnit\Framework\Attributes\After;
+use PHPUnit\Framework\Attributes\Before;
 
 /**
  * Integrates Mockery into PHPUnit. Ensures Mockery expectations are verified
@@ -62,6 +64,7 @@ trait MockeryPHPUnitIntegration
     /**
      * @before
      */
+    #[Before]
     protected function startMockery()
     {
         $this->mockeryOpen = true;
@@ -70,6 +73,7 @@ trait MockeryPHPUnitIntegration
     /**
      * @after
      */
+    #[After]
     protected function purgeMockeryContainer()
     {
         if ($this->mockeryOpen) {


ghostwriter (approved):
Looks good to me, @ruudk

Thank you.

ghostwriter:
GitHub is partially down right now,

"degraded performance for Pull Requests and Actions:"  https://www.githubstatus.com/incidents/wcl1sw4mzg60

I’ll restart GitHub Action and merge this once CI completes.

ghostwriter:
for the record...

> When both `/** @before */` and `#[Before]` are used then **1)** the before-test method is called only once and **2)** the deprecation is not emitted.

https://github.com/sebastianbergmann/phpunit/issues/5695#issuecomment-1929436000

westy92:
Thanks for fixing this! Looking forward to the release. 🚀

PR #1397, opened 2024-03-13
fix mock method $args parameter type

### LINKED ISSUE #1395: PHPStan error with mock method in 1.6.8
### Mockery Version

1.6.8

### PHP Version

PHP 8.3

### Issue Description

phpstan does not detect return value anymore properly

### Steps to Reproduce

```php

class Test
{
    public function test() : string
    {
        return 'Test';
    }
}

$test = \Mockery::mock(Test::class);

$test->shouldReceive('test')->andReturn('Mocked');

echo $test->test();

```

### Expected Behavior

No PHPStan errors

### Actual Behavior

phpstan reports now following error: phpstan: Return type of call to static method Mockery::mock() contains unresolvable type.

### Exception or Error

_No response_

### Additional Information

_No response_

sakarikl:
fix for #1395

### DIFF
--- a/library/Mockery.php
+++ b/library/Mockery.php
@@ -418,7 +418,7 @@ class Mockery
      *
      * @template TInstanceMock
      *
-     * @param TInstanceMock ...$args
+     * @param array<class-string<TInstanceMock>|TInstanceMock> $args
      *
      * @return LegacyMockInterface&MockInterface&TInstanceMock
      */
@@ -468,7 +468,7 @@ class Mockery
      *
      * @template TMock
      *
-     * @param TMock ...$args
+     * @param array<class-string<TMock>|TMock> $args
      *
      * @return LegacyMockInterface&MockInterface&TMock
      */
@@ -496,7 +496,7 @@ class Mockery
      *
      * @template TNamedMock
      *
-     * @param TNamedMock ...$args
+     * @param array<class-string<TNamedMock>|TNamedMock> $args
      *
      * @return LegacyMockInterface&MockInterface&TNamedMock
      */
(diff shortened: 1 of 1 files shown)


sakarikl:
This still does not fix alias: and such usages, but most normal use cases PHPStan undestands it properly.

and if no parameters given at all also still fails.


But in those cases it might be able to solve with this: https://github.com/phpstan/phpstan-mockery

ghostwriter (approved):
Looks great, going to merge this pull request.

Thank you.

ghostwriter:
> This still does not fix alias: and such usages

Thank you, made a note to take care of that.

library/Mockery.php
```diff
@@ -468,7 +468,7 @@ public static function isSame($expected): IsSame
      *
      * @template TMock
      *
-     * @param TMock ...$args
+     * @param array<class-string<TMock>|TMock> $args
```
LastDragon-ru, line 471: This is invalid type, because according in the doc you can

```
```

sakarikl: Yes it is. There are still a lot of different use cases that this does not cover.

Harrald:
While this update fixes a lot of phpstan errors. It also introduces a new one for an edge case in my code:

`Parameter #2 ...$args of static method Mockery::mock() expects class-string<SplFileObject>|SplFileObject, array<int, string> given. `
```
$path = 'php://memory';
$splFileObject = Mockery::mock('SplFileObject', [$path]);
```

It seems that like it only excepts 1 parameter?
These also fail:
`$skuNotFoundException = Mockery::mock('alias:' . SkuDoesNotExistException::class, Exception::class);`
`$supplierModel = Mockery::mock('alias:' . SupplierModel::class, Model::class);`

PR #1401, opened 2024-03-20
Improve `Mockery::mock()` parameter type compatibility with array typehints

### LINKED ISSUE #1400: Invalid `Mockery::mock()`/`spy()` phpdoc (= phpstan errors)
### Mockery Version

1.6.10

### PHP Version

PHP 8.3

### Issue Description

Today I've updated to the latest 1.6.10 and see a lot of phpstan errors... Seems they were introduced in #1397 where the type of `$args` was changed to `@param array<class-string<TMock>|TMock> $args`. After this change, any of

```
$mock = \Mockery::mock('MyClass', [$constructorArg1, $constructorArg2]);
```

will generate phpstan error

```
 463    Parameter #2 ...$args of static method Mockery::mock() expects class-string<Directive>|Directive, array<int, StreamFactory> given.
  716    Parameter #2 ...$args of static method Mockery::mock() expects class-string<Directive>|Directive, array<int, StreamFactory> given.
  1142   Parameter #2 ...$args of static method Mockery::mock() expects class-string<Directive>|Directive, array<int, StreamFactory> given.
  1236   Parameter #2 ...$args of static method Mockery::mock() expects class-string<Directive>|Directive, array<int, StreamFactory&Mockery\MockInterface> given.    
  1273   Parameter #2 ...$args of static method Mockery::mock() expects class-string<Directive>|Directive, array<int, StreamFactory&Mockery\MockInterface> given.    
```

Maybe something like will be better (but seems it is a breaking change and will break `phpstan/phpstan-mockery`) 

```
/**
 * @template TMock
 *
 * @param array<class-string<TMock>|TMock> $class
 *
 * @return LegacyMockInterface&MockInterface&TMock
 */
public static function mock(mixed $class, mixed ...$args)
```

Also, the `Mockery::spy(static function () {})` now gives (not sure related to `phpstan/phpstan-mockery` or not):

```
Return type of call to static method Mockery::spy() contains unresolvable type.
```

**PS:** Original issue (#1395) probably can be solved with `phpstan/phpstan-mockery` without changing docblock (at least I have used it for a while and never see any issue)

### Steps to Reproduce

Update to 1.6.10 & run phpstan

### Expected Behavior

No new phpstan errors

### Actual Behavior

A lot of new phpstan errors

### Exception or Error

_No response_

### Additional Information

_No response_

LastDragon-ru:
Seems fixes #1400 and original #1395/#1397, except for `Mockery::mock()` and `spy(function() {})`. Would be nice if someone can test it too (because I use `mock()` and `spy()` only)

@sakarikl

### DIFF
--- a/library/Mockery.php
+++ b/library/Mockery.php
@@ -418,7 +418,7 @@ class Mockery
      *
      * @template TInstanceMock
      *
-     * @param array<class-string<TInstanceMock>|TInstanceMock> $args
+     * @param array<class-string<TInstanceMock>|TInstanceMock|array<mixed>> $args
      *
      * @return LegacyMockInterface&MockInterface&TInstanceMock
      */
@@ -468,7 +468,7 @@ class Mockery
      *
      * @template TMock
      *
-     * @param array<class-string<TMock>|TMock> $args
+     * @param array<class-string<TMock>|TMock|array<mixed>> $args
      *
      * @return LegacyMockInterface&MockInterface&TMock
      */
@@ -496,7 +496,7 @@ class Mockery
      *
      * @template TNamedMock
      *
-     * @param array<class-string<TNamedMock>|TNamedMock> $args
+     * @param array<class-string<TNamedMock>|TNamedMock|array<mixed>> $args
      *
      * @return LegacyMockInterface&MockInterface&TNamedMock
      */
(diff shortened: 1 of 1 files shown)


aldilop (approved):
I can confirm this change fixes it for me, tho in my code I'm only using `Mockery::mock`
Anyway it makes sense for the rest of the methods.

sakarikl:
This works in all cases as the previous one worked for us. And fixed cases where constructor parameters were given in array as second parameter.

Still does not work with 'alias:' . $object::class type of parameters as did not the previous one.

ghostwriter (approved):
Thank you for the improvements @LastDragon-ru,

- Made a note to address the `alias:` and `overload:`.

Thank you both for the feedback @sakarikl & @aldilop

PR #1403, opened 2024-03-20
Fix regression - partial construction with trait methods

### LINKED ISSUE #1402: Issues with partial construction trait methods on 1.6.10 - change in behaviour
### Mockery Version

1.6.10

### PHP Version

PHP 8.1

### Issue Description

I updated from 1.6.9 to 1.6.10 this morning, and a number of my tests have started failing. We have a number of (legacy!) traits that do different things, some of which have properties that need initialising when the class is created, so the trait has an initialise method that we call in the constructor.

If we want to create a partial mock of one of these classes, where we pass in constructor arguments (and so call the constructor), and want to override a different method on the class for some reason, we can no longer do that on 1.6.10.

### Steps to Reproduce

I made a minimal example! 

https://github.com/adnweedon/mockeryIssue

This will pass if you run composer run test, but if you update to 1.6.10 the test will start failing.

### Expected Behavior

It makes sense to me that a mocked class (ie, all the methods are mocked, we haven't made it partial yet) would have issues with constructor methods, as we've not defined expectations for the method yet - it just wasn't previously the behaviour! I don't understand why the changes on 1.6.10 are causing the issue, but I would expect a patch update not to break a whole bunch of tests.

### Actual Behavior

A number of my tests are now failing, because when Mockery calls the constructor, it's trying to find a mocked version of the trait method, which an expectation hasn't been set for.

### Exception or Error

_No response_

### Additional Information

_No response_

ghostwriter:
This patch resolves #1402, via the following changes:

 - [Call get_parent_class before calling the parent class](https://github.com/mockery/mockery/commit/61a65ce03211d93c1a9681ce514c6aa9aefbfcde)

### DIFF
--- a/library/Mockery.php
+++ b/library/Mockery.php
@@ -418,7 +418,7 @@ class Mockery
      *
      * @template TInstanceMock
      *
-     * @param array<class-string<TInstanceMock>|TInstanceMock> $args
+     * @param array<class-string<TInstanceMock>|TInstanceMock|array<mixed>> $args
      *
      * @return LegacyMockInterface&MockInterface&TInstanceMock
      */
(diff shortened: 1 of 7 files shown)


adnweedon:
Thank you!!

PR #1410, opened 2024-03-22
Fix PHP Doc Comments

### LINKED ISSUE #1411: Mocking final classes reports unresolvable type by PHPStan
### Mockery Version

1.6.11

### PHP Version

PHP 8.2

### Issue Description

PHPStan returns `Return type of call to static method Mockery::mock() contains unresolvable type.` when I try to mock a class that is `final` (in this case it's also readonly, but the issue is caused by the final statement).

```php
$dto = \Mockery::mock(MyFinalClass::class);
```

In my PHPUnit bootstrap I setup [dg\bypass-finals](https://github.com/dg/bypass-finals) so it "technically" possible to create the mock. This has always worked up to and including version `1.6.9`. It's only the last 2 releases of Mockery that PHPStan started reporting the above errors.

### Steps to Reproduce

1. Create a final class.
2. Try and mock the final class in a PHPUnit test with `dg/bypass-finals` configured in PHPUnit's bootstrap file.
3. Run PHPStan.

### Expected Behavior

No errors PHPStan errors

### Actual Behavior

PHPStan reports on the line with `Mockery::mock()`:

```
 Return type of call to static method Mockery::mock() contains unresolvable type.
```

### Exception or Error

_No response_

### Additional Information

_No response_

rolandsusans:
### Related issue:
- https://github.com/mockery/mockery/issues/1400

### Steps to reproduce:

When installing the latest:
- `mockery/mock:1.6.11` in project dependencies, the related issue persists when Closure used.

Example code that generates the phpstan issue:
```php
Mockery::mock(
    LoggerInterface::class,
    static function (MockInterface $mock): void {
       //....
})
```

`phpstan` error:
```
-
	message: "#^Parameter \\#2 \\.\\.\\.\\$args of static method Mockery\\:\\:mock\\(\\) expects array\\|class\\-string\\<Interop\\\\Queue\\\\Consumer\\>\\|Interop\\\\Queue\\\\Consumer, Closure given\\.$#"
	count: 8
	path: tests/Unit/.......php
```
### Description 
This pull request includes a change to the `mock` function in the `Mockery.php` file. The change expands the type of the `$args` parameter to also include `Closure` along with the existing types.

This change allows `Closure` objects to be passed as arguments to the `mock` function, which fixes mentioned phpstan error.


aldilop:
Somehow related, I think the phpdoc in `namedMock` is also broken.
Phpstan should not be able to guess `TNamedMock` if for example we:
```php
$mock = Mockery::namedMock('AnotherName', MyClass::class);
```

ghostwriter:
there are many ways to create a mock, this list may not be complete.

```php
$mock = \Mockery::mock();

$mock = \Mockery::mock(MyClass::class);

$mock = \Mockery::mock(MyClass::class, function ($mock) {
    $mock->shouldAllowMockingProtectedMethods();
});

$mock = \Mockery::mock(array('pi' => 3.1416, 'itWorks' => true));
$this->assertEquals(3.1416, $mock->pi());

$mock = \Mockery::mock('MyClass, MyInterface');

$mock = \Mockery::mock('MyClass', 'MyInterface, OtherInterface');

// <!-- runtime partial test doubles, -->
class Foo {
    function foo() {
       return 123;
    }
    function bar() {
        return $this->foo(); 
    }
}
$foo = mock(Foo::class)->makePartial();
$foo->foo(); // int(123);

// <!-- generated partial test doubles, and -->
$mock = \Mockery::mock('MyClass[method]', [param]);

// <!-- proxied partial test doubles. -->
$mock = \Mockery::mock(new MyClass);

// <!-- aliasing, -->
$mock = \Mockery::mock('alias:MyClass');

// <!-- and overloading. -->
$mock = \Mockery::mock('overload:MyClass');

// <!-- named mocks, -->
$mock = \Mockery::namedMock('MyClassName', 'DateTime');

// <!-- with constructor arguments, -->
$mock = \Mockery::mock('MyClass', [$constructorArg1, $constructorArg2]);

// <!-- There’s also a shorthand way of setting up method call expectations and their return values: -->
$mock = \Mockery::mock('MyClass', ['name_of_method_1' => 'return value 1', 'name_of_method_2' => 'return value 2']);

// <!-- In the next example, the foo() and bar() methods of MyClass will be mocked but no other MyClass methods are touched.
// We will need to define expectations for the foo() and bar() methods to dictate their mocked behaviour. -->
$mock = \Mockery::mock('MyClass[foo,bar]');

// <!-- pass in constructor arguments since unmocked methods may rely on those -->
$mock = \Mockery::mock('MyNamespace\MyClass[foo]', array($arg1, $arg2));

```

we’ll add tests using each syntax, including `spy|namedMock|mock` and run static analysis tools, with their plugins.

aldilop:
Everything that can be done with `mock` can be done with `namedMock`, it just shifts the first parameter.
There's a lot more magic in `mock` that I knew of tho.

ghostwriter:
Im going to merge this do all of the additional work in a separate PR.

library/Mockery/CountValidator/AtMost.php
```diff
@@ -40,5 +41,7 @@ public function validate($n)
                 ->setActualCount($n);
             throw $exception;
         }
+
+        return true;
```
ghostwriter:
```suggestion
```

library/Mockery/CountValidator/AtLeast.php
```diff
@@ -45,12 +46,15 @@ public function validate($n)
...
+
             $exception->setMock($this->_expectation->getMock())
                 ->setMethodName((string) $this->_expectation)
                 ->setExpectedCountComparative('>=')
                 ->setExpectedCount($this->_limit)
                 ->setActualCount($n);
             throw $exception;
         }
+
+        return true;
```
ghostwriter:
```suggestion
```

library/Mockery/CountValidator/Exact.php
```diff
@@ -43,5 +44,7 @@ public function validate($n)
                 ->setActualCount($n);
             throw $exception;
         }
+
+        return true;
```
ghostwriter:
```suggestion
```

PR #1450, opened 2025-01-30
Phpunit integration test listener false positive

steefmin:
This change proposes resetting the container when the testSuite is started. This fixes a false positive that is thrown by the TestListener when a mock is created before running the actual tests. 

This might require some context: We sometimes create mocks in PHPunit's dataproviders. This might not be intended, but for some of our tests it makes sense. When PHPunit runs, it will execute all dataproviders to determine all the test cases to run (for progress counts and filtering test cases). When mocks are created in the dataprovider, the mockery container is initialized before the actual tests are run. Afterward the test listener will assert that the container is closed on end of a test. If the first test that is executed is a test that does not use Mockery and thus does not close it, the testListener will issue a false possitive failure that Mockery's expectations are not verified. 

It can be summarized: 
- phpunit is started
- phpunit runs dataproviders
- dataprovider creates a mock
- Mockery container is created
- phpunit starts suite
- phpunit runs non-mockery test
- non-mockery test does not close Mockery
- non-mockery test ends
- phpunit triggers listener
- listener detects Mockery is not closed
- listener falsely reports failure

This change fixes it by resetting the container between creating mocks in data providers and running initial test

notes: 
- I had to remove some pre-condition checks in `\Tests\Unit\Mockery\Adapter\Phpunit\TestListenerTest::testMockeryIsAddedToBlacklist` as `\Mockery\Adapter\Phpunit\TestListener::startTestSuite` is now called multiple times in tests. It sets a singleton in PHPUnit, thus to keep tests independent I had to remove it. (other option would be to reset the internals of `\PHPUnit\Util\ExcludeList` using reflection, but that seems unwanted as it might change unexpectedly)

### DIFF
--- a/library/Mockery/Adapter/Phpunit/TestListenerTrait.php
+++ b/library/Mockery/Adapter/Phpunit/TestListenerTrait.php
@@ -80,5 +80,6 @@ class TestListenerTrait
         } else {
             Blacklist::$blacklistedClassNames[Mockery::class] = 1;
         }
+        Mockery::resetContainer();
     }
 }
--- a/tests/Unit/Mockery/Adapter/Phpunit/TestListenerTest.php
+++ b/tests/Unit/Mockery/Adapter/Phpunit/TestListenerTest.php
@@ -43,7 +43,7 @@ final class TestListenerTest extends MockeryTestCase
     {
         $this->listener->startTestSuite(new TestSuite());
 
-        $mock = Mockery::getContainer()->mock();
+        $mock = Mockery::mock();
         $mock->shouldReceive('bar')
             ->once();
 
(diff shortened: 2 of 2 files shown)


steefmin:
I see that the workflow step for Psalm has failed. But I don't see how these errors came from the proposed changes, nor do I see improvements for the code by "fixing" them. Fixing in quotes here as most of them would require me to add some annotations or add the errors to the baseline, both solutions do not feel helpfull to me. Please let me know what you expect me to do with them.

ghostwriter (approved):
Hey @steefmin,

This PR looks great, I'll merge it and handle the Psalm issue in a separate PR.

> I see that the workflow step for Psalm has failed. But I don't see how these errors came from the proposed changes, nor do I see improvements for the code by "fixing" them. Fixing in quotes here as most of them would require me to add some annotations or add the errors to the baseline, both solutions do not feel helpfull to me. Please let me know what you expect me to do with them.

Yeah, I agree. No need to worry about them; I’ll address them separately.

I just need to pull and run the end-to-end tests locally, even though this only affects PHPUnit 9.6.x users, should be all good. 

Thanks for your work on this!

PR #1451, opened 2025-02-02
Add `withAndOthers` expectation shorthand

bert-w:
I want to request to add the `withAndOthers` shorthand function so people can more easily write expectations for (changing) optional arguments.

Usage:
```php
public function foobar(string $foo, int $bar, array $args = []): void
{
    // ...
}

$mock->shouldReceive('foobar')->withAndOthers('str', 123);
```

Use case: when I add a new optional parameter to a function, some testcases that use `$mock->with(...)` will fail (since a new parameter might now be included), however by using `$mock->withAndOthers(...)` I can indicate that the optional parameters have no expectation and will be accepted in any form.

I don't know if this requires a test since there is one already which tests the underlying logic:

https://github.com/mockery/mockery/blob/dca1e95368d49db1613998c2c1dfd5311e3c8f75/tests/Unit/Mockery/ExpectationTest.php#L139-L145

### DIFF
--- a/library/Mockery/Expectation.php
+++ b/library/Mockery/Expectation.php
@@ -797,7 +797,7 @@ class Expectation implements ExpectationInterface
 
     public function withAndOthers(...$args)
     {
-        return $this->withArgs([...$args, new AndAnyOtherArgs()]);
+        return $this->withArgs(array_merge($args, [new AndAnyOtherArgs()]));
     }
 
     /**
--- a/library/Mockery/Expectation.php
+++ b/library/Mockery/Expectation.php
@@ -788,6 +788,19 @@ class Expectation implements ExpectationInterface
     }
 
     /**
+     * Set expectation that the given arguments are acceptable plus variable other arguments
+     *
+     * @param mixed ...$args
+     *
+     * @return self
+     */
+
+    public function withAndOther(...$args)
+    {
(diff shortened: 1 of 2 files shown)


ghostwriter (changes requested):
Hey @bert-w,

This looks good—thanks!

> I don't know if this requires a test since there is one already which tests the underlying logic:

Yes, a test is needed since we're introducing a new alternative syntax.

Can you copy the existing test you mentioned and rename it to `testAndAnyOtherConstraintMatchesTheRestOfTheArgumentsAlternativeSyntax`?

Then, update it to use the `withAndOthers` method you introduced.

Appreciate your help!

bert-w:
i've changed the syntax to use php 7.3 (didnt notice this repo was still on that version) + added a test similar to existing ones (I havent tested this locally since I didn't set it up).

bert-w:
I cant run the Psalm tool locally (windows). Honestly I question the Psalm usage for this repository when the baseline has 6000+ exclusions :|

ghostwriter:
> I cant run the Psalm tool locally (windows). Honestly I question the Psalm usage for this repository when the baseline has 6000+ exclusions :|

No need to worry about them; I’ll address the Psalm issue in a separate PR.

Thanks

ghostwriter (approved):

PR #1460, opened 2025-04-10
Fixes CVE-2025-27516

dustin-scott:
Fixes https://nvd.nist.gov/vuln/detail/CVE-2025-27516 which show up because of the python dependencies in the docs.
We may end up whitelisting since we don't use python personally, but figured this would reduce some pain.


ghostwriter:
Thank you @dustin-scott

PR #1463, opened 2025-05-22
Docs: add alternative approaches for mocking final classes

amyavari:
I know mentioning other packages in Mockery's documentation might not be appropriate, but in this case, I think it's a generous and helpful gesture for users.

Since Mockery is the most popular mocking library in PHP, many developers treat its documentation as the single source of truth for mocking dependencies in tests. So when Mockery says that for final classes there's almost nothing that can be done and suggests changing the design, most users don’t look for other solutions.

But when developers want to write strict code—marking classes as `final`, type-hinting properly, and using the full features of their frameworks like Laravel’s service container—they often face tough decisions.

So I thought it might be a good idea to briefly introduce packages that can remove the final keyword on the fly. This way, users can stay strict, enjoy their framework’s features, and still use Mockery with ease.

Thank you for considering this.


ghostwriter:
Thanks @amyavari

PR #1474, opened 2025-11-16
Mock demeter chains and fluent interface methods with `static` return type

### LINKED ISSUE #1447: Mocking an interface method with a static return value as part of a demeter chain mock results in a fatal error
### Mockery Version

1.6.12

### PHP Version

PHP 8.2

### Issue Description

As per the title - if you create an interface with a method that has a static return type, Mockery fails.

### Steps to Reproduce

1. Create an interface
2. Add a method that has a static return type
3. Run test

### Expected Behavior

The method with the static return type should be mockable.

### Actual Behavior

Mockery throws an exception:

```cli
syntax error, unexpected token "static", expecting identifier

  at /tmp/MockeryVc3BMx:1
  ➜   1▕ <?php class static {}

  1   /tmp/MockeryVc3BMx:1
  2   app/Modules/Forms/Formables/RequestToFieldValues.test.php:24
```

### Exception or Error

Full error is above.

### Additional Information

_No response_

kirkbushell:
As per the title, this isn't a solution to #1447 but instead illustrates the bug with a new test case.

We use the static return type for interfaces, as self can result in child classes not being able to work with said interface.

In short, if you have an interface method signature which has a static return type, mockery fails:

```cli
ParseError: syntax error, unexpected token "static", expecting identifier
```

The test case added shows this aspect failing. I've tried solving this but honestly, I am struggling to follow the mockery code, so any guidance would be appreciated and I'll update this PR to solve the problem appropriately.

Note that this can still be done with a hacky utility method we wrote as a wrapper to help us still use demeter chain expectations on interfaces:

```php
protected function demeter(MockInterface $mock, string $chain): HigherOrderMessage|Expectation|ExpectationInterface
{
    $expectations = explode('->', $chain);
    $final = array_pop($expectations);

    $mock->shouldReceive(Arr::mapWithKeys($expectations, fn($expectation) => [$expectation => $mock]));

    return $mock->shouldReceive($final);
}
```

And it can also be done using array expectations:

```php
$mock->shouldReceive(['do' => $mock, 'thing' => $mock])->andReturn(...);
```

So it seems to me that something funky is going on specifically with interfaces deeper down in the mockery code. Ideally, we want to be able to support demeter chain mocks on interface methods with static return types.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -44,7 +44,7 @@
         "security": "https://github.com/mockery/mockery/security/advisories"
     },
     "require": {
-        "php": ">=7.3",
+        "php": ">=7.3|>=8.0",
         "lib-pcre": ">=7.0",
         "hamcrest/hamcrest-php": "^2.0.1"
     },
--- a/composer.json
+++ b/composer.json
@@ -44,7 +44,7 @@
         "security": "https://github.com/mockery/mockery/security/advisories"
     },
     "require": {
-        "php": ">=7.3|>=8.0",
+        "php": ">=7.3",
         "lib-pcre": ">=7.0",
         "hamcrest/hamcrest-php": "^2.0.1"
     },
--- a/library/Mockery.php
+++ b/library/Mockery.php
@@ -13,6 +13,7 @@ use Mockery\CompositeExpectation;
 use Mockery\Configuration;
 use Mockery\Container;
 use Mockery\Exception as MockeryException;
+use Mockery\ExpectationDirector;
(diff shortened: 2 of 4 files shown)


ghostwriter:
Hey @kirkbushell,

Thank you for your contribution. 

I appreciate the time you invested in preparing this pull request.

I was able to address the issue with a few additional commits.

---

Syntax:

`$mock->shouldReceive('do->thing');` or `$mock->shouldReceive('do->thing')->andReturnSelf();`

PR #1483, opened 2026-05-05
Add export-ignore rules to .gitattributes

sasezaki:
Current archive targets other than library/ will be as follows:
```bash
gh api repos/mockery/mockery/tarball/HEAD | tar -tzf - | cut -d/ -f2- | sed '/^$/d' | grep -vE '^library/' | sort
```

```
CHANGELOG.md
composer.json
composer.lock
CONTRIBUTING.md
COPYRIGHT.md
LICENSE
.phpstorm.meta.php
README.md
.readthedocs.yml
SECURITY.md
```

With this PR, the archive targets other than library/ will be as follows:
```bash
gh api repos/sasezaki/mockery/tarball/patch-gitattributes | tar -tzf - | cut -d/ -f2- | sed '/^$/d' | grep -vE '^library/' | sort
```

```
CHANGELOG.md
composer.json
CONTRIBUTING.md
COPYRIGHT.md
LICENSE
.phpstorm.meta.php
README.md
SECURITY.md
```

Changes:
- Added `/composer.lock export-ignore`
- Added `/.readthedocs.yml export-ignore`
---
> **Note:** `composer.lock` is included in dist archives by default, but for libraries it is unnecessary — Composer ignores `composer.lock` from dependencies and resolves versions from `composer.json` alone. Excluding it keeps the dist archive clean.
> **Personal opinion:** `CONTRIBUTING.md` exists in this repository. I personally consider it unnecessary in the dist archive, but I have chosen not to add it to `export-ignore`.

PR #1486, opened 2026-05-20
Update .gitattributes to exclude dev files from composer dist

williamdes:
This avoids shipping dev/tooling files in the composer dist that aren't needed at runtime.

Last `.gitattributes` update was daad681 (2026-05-05, #1483). The files below have been added or modified since, and are still included in the composer dist:
- `/CONTRIBUTING.md export-ignore` — last touched in 5ee21d6 2023-12-10

Background reading: https://blog.madewithlove.be/post/gitattributes/

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -19,3 +19,4 @@
 /tools export-ignore
 /composer.lock export-ignore
 /.readthedocs.yml export-ignore
+/CONTRIBUTING.md export-ignore


ghostwriter:
Hi @williamdes,

No, thank you.

We have already addressed this in, https://github.com/mockery/mockery/pull/1483, as you've mentioned.

williamdes:
Hi @ghostwriter 
Okay I understand that you prefer to keep the file.
I have doubts about `.phpstorm.meta.php`
Do you also want to keep it ?

PR #1495, opened 2026-07-06
Do not return from constructors/destructors

### LINKED ISSUE #1494: Dynamic __construct code contains deprecated `return $ret;` call
### Mockery Version

1.6.12

### PHP Version

PHP 8.6 (alpha 1)

### Issue Description

Returning a value from a constructor is deprecated in PHP 8.6; see also https://wiki.php.net/rfc/deprecate-return-value-from-construct.

When creating a mock with dynamic code, I observe the following deprecation:
```
ErrorException: Returning a value from a constructor is deprecated

/app-code/vendor/mockery/mockery/library/Mockery/Loader/EvalLoader.php:30
/app-code/vendor/mockery/mockery/library/Mockery/Loader/EvalLoader.php:30
/app-code/vendor/mockery/mockery/library/Mockery/Container.php:405
/app-code/vendor/mockery/mockery/library/Mockery.php:477
```

After some debugging, I found the following snippet being processed in `EvalLoader`:
```
    public function __construct($app){\n
    $argc = func_num_args();\n
    $argv = func_get_args();\n
    $ret = $this->_mockery_handleMethodCall(__FUNCTION__, $argv);\n
    return $ret;\n
    }\n
```

Likely somewhere along the following code an explicit exclusion for `return` should be added for constructors:
https://github.com/mockery/mockery/blob/3a80322e874fbdce4e87e739456fe48d48a527c8/library/Mockery/Generator/StringManipulation/Pass/MethodDefinitionPass.php#L193

### Steps to Reproduce

_No response_

### Expected Behavior

_No response_

### Actual Behavior

_No response_

### Exception or Error

_No response_

### Additional Information

_No response_

lazerg:
Returning a value from a constructor is deprecated as of PHP 8.6, so the generated `__construct` no longer appends `return $ret;`. The handler is still called, so constructor expectations keep working.

Fixes #1494.

### DIFF
--- a/.github/workflows/psalm.yml
+++ b/.github/workflows/psalm.yml
@@ -24,9 +24,9 @@ jobs:
         uses: shivammathur/setup-php@v2
         with:
           coverage: none
-          php-version: 8.4
+          php-version: 8.5
           ini-values: assert.exception=1, zend.assertions=1, error_reporting=-1, log_errors_max_len=0, display_errors=On
-          tools: psalm:6.7.1
+          tools: psalm:6.16.1
 
       - name: Get composer cache directory
         id: composer-cache
--- a/library/Mockery/Generator/StringManipulation/Pass/MethodDefinitionPass.php
+++ b/library/Mockery/Generator/StringManipulation/Pass/MethodDefinitionPass.php
@@ -72,11 +72,13 @@ class MethodDefinitionPass implements Pass
     protected function renderParams(Method $method, $config)
     {
         $class = $method->getDeclaringClass();
+        $className = strtolower($class->getName());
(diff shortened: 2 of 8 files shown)


jnoordsij:
Thanks a lot for the quick efforts here! FWIW I found the following reference where a similar change is applied to PHPUnit: https://github.com/sebastianbergmann/phpunit/commit/21673a4506bd831939bc4510d0ef4b592cf734b9.

ghostwriter:
Thanks for your help, @lazerg.

LordSimal:
Hey @ghostwriter, good to see you here.

What are your plans for a new release for mockery? Is there something you need help with?

PR #1509, opened 2026-08-16
Fix str_* polyfills documenting non-empty-string parameters

### LINKED ISSUE #1510: 1.6.13 polyfills annotate str_contains/str_starts_with/str_ends_with with non-empty-string, breaking static analysis on PHP 8
library/helpers.php in 1.6.13 declares polyfills for str_contains(), str_starts_with() and str_ends_with() with @param non-empty-string $haystack and @param non-empty-string $needle.

That contract is narrower than PHP's own (both parameters are plain string), and it is also wrong for the polyfill bodies themselves — they explicitly handle empty strings ('' === $needle returns true, '' === $haystack returns false).

The bigger problem is the effect on consumers. library/helpers.php is a Composer autoload.files entry, so static analysers index those declarations regardless of the function_exists() guard, and a userland declaration overrides the analyser's native stub. On PHP 8.5, where the polyfills never actually run, upgrading 1.6.12 → 1.6.13 produced ~380 new PHPStan (level 9) errors across our codebase:

Parameter #1 $haystack of function str_contains expects non-empty-string, string given.
Suggested fix: drop the non-empty-string annotations (use string), or guard the polyfill declarations so they aren't indexed on PHP 8+ (e.g. move them into a separate file only required on PHP < 8).

Reproduce: a PHP 8 project with mockery 1.6.13 and PHPStan, any str_contains($a, $b) call where both are string.

dg:
The polyfills for str_contains(), str_ends_with() and str_starts_with() documented both parameters as non-empty-string. That contradicts the native functions as well as the polyfill bodies, which all handle an empty string.

Since library/helpers.php is autoloaded through composer, static analysers read these declarations as the definition of the global functions, so any project requiring Mockery started to report:

    Parameter #1 $haystack of function str_starts_with expects
    non-empty-string, string given.

### DIFF
--- a/library/helpers.php
+++ b/library/helpers.php
@@ -137,9 +137,6 @@ if (! \function_exists('str_contains')) {
      *
      * @copyright Fabien Potencier
      * @license https://github.com/symfony/polyfill/blob/1.x/src/Php80/LICENSE
-     *
-     * @param non-empty-string $haystack
-     * @param non-empty-string $needle
      */
     function str_contains(string $haystack, string $needle): bool
     {
@@ -153,9 +150,6 @@ if (! \function_exists('str_ends_with')) {
      *
      * @copyright Fabien Potencier
      * @license https://github.com/symfony/polyfill/blob/1.x/src/Php80/LICENSE
-     *
-     * @param non-empty-string $haystack
-     * @param non-empty-string $needle
      */
     function str_ends_with(string $haystack, string $needle): bool
     {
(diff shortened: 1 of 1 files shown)


jnoordsij:
Only just spotted this after creating #1511, which is an alternative approach to the problem at hand, which is reported in #1510.

ghostwriter:
@dg, Thanks David!

